PR #4, opened 2018-04-16
Static index

digeratus:
Here's the PR for the static index sample using a directory and default redirect file.


fafhrd91:
@digeratus could you remove license file?

digeratus:
I just removed the LICENSE file.

fafhrd91:
thanks!

PR #15, opened 2018-05-31
Listen to localhost instead of all network interfaces

caulagi:
When we listen on all network interfaces instead of localhost, Mac shows this pop window. Maybe listening on 0.0.0.0 was not intended?

screen shot 2018-05-31 at 21 51 56


fafhrd91:
thanks!

PR #16, opened 2018-05-31
Removed duplicate dependency in askama Cargo.toml

caulagi:
It seems to me that the build dependency section is not required, since it is covered under dependencies. Or am I missing something?


fafhrd91:
thanks! I don't remember where this sections came from, might be just copy/paste.

PR #19, opened 2018-06-15
Added example with very fast compile-time template engine Maud

max-frai:
Recently maud merged responder for their markup type and now we can return template result directly from handler.
Maud is probably the fastest template engine now with very powerful features.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,6 +52,7 @@ script:
       cd static_index && cargo check && cd ..
       cd template_askama && cargo check && cd ..
       cd template_tera && cargo check && cd ..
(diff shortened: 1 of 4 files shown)


DoumanAsh:
I would prefer not to depend on git dependencies.
So let's wait for maud actual release with actix support

max-frai:
Sure, I agree with you, also I've updated travis config.
When changes will be merged into master I will update this ticket.

fafhrd91:
@max-frai you should run maud only on nightly

DoumanAsh:
@max-frai ping?

fafhrd91:
outdated

PR #21, opened 2018-06-20
Add state to the multipart example

xvny:
Add state to the multipart example

### DIFF
--- a/multipart/src/main.rs
+++ b/multipart/src/main.rs
@@ -6,6 +6,7 @@ extern crate futures;
 
 use std::fs;
 use std::io::Write;
+use std::cell::Cell;
 
 use actix_web::{
     error, http, middleware, multipart, server, App, Error, FutureResponse, HttpMessage,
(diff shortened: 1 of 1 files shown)


multipart/src/main.rs
```diff
@@ -15,9 +16,11 @@ use actix_web::{
 use futures::future;
 use futures::{Future, Stream};
 
-pub fn save_file(
-    field: multipart::Field<HttpRequest>,
-) -> Box<Future<Item = i64, Error = Error>> {
+pub struct AppState {
+    pub counter: Cell<usize>,
```
DoumanAsh: Why to use `Cell`?

multipart/src/main.rs
```diff
@@ -15,9 +16,11 @@ use actix_web::{
...
+}
+
+pub fn save_file( field: multipart::Field<HttpRequest<AppState>>) -> Box<Future<Item = i64, Error = Error>> {
```
DoumanAsh: Could you please remove leading space before `field`

multipart/src/main.rs
```diff
@@ -42,9 +45,7 @@ pub fn save_file(
     )
 }
 
-pub fn handle_multipart_item(
-    item: multipart::MultipartItem<HttpRequest>,
-) -> Box<Stream<Item = i64, Error = Error>> {
+pub fn handle_multipart_item( item: multipart::MultipartItem<HttpRequest<AppState>>) -> Box<Stream<Item = i64, Error = Error>> {
```
DoumanAsh: Could you please remove leading space before `item`

multipart/src/main.rs
```diff
@@ -57,7 +58,9 @@ pub fn handle_multipart_item(
     }
 }
 
-pub fn upload(req: HttpRequest) -> FutureResponse<HttpResponse> {
+pub fn upload(req: HttpRequest<AppState>) -> FutureResponse<HttpResponse> {
+    req.state().counter.set(req.state().counter.get() + 1);
+    println!("{:?}",  req.state().counter.get());
```
DoumanAsh: I feel like there is not much need to print just counter

Well maybe you could just print something `Upload number: {}`?

PR #25, opened 2018-07-04
typos -> text/plain

casperin:
- Change `plain/text` to `text/plain` in [basics/README.md](https://github.com/actix/examples/blob/master/basics/README.md)
- Change `test/plain` to `text/plain` in [basics/src/main.rs](https://github.com/actix/examples/blob/master/basics/src/main.rs)


DoumanAsh:
Thank you

PR #26, opened 2018-07-04
Form example

### LINKED ISSUE #24: form data example

casperin:
Examples on how to handle form submits

Closes #24

### DIFF
--- a/form/Cargo.toml
+++ b/form/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "form-example"
+version = "0.1.0"
+authors = ["Gorm Casper <gcasper@gmail.com>"]
+
+[dependencies]
+actix = "0.5"
+actix-web = "^0.6"
+serde = "1.0.69"
+serde_derive = "1.0.69"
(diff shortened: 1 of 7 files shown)


casperin:
Not specifically, but vim just does that for me. Like it did for https://github.com/actix/actix-web/pull/362

DoumanAsh:
It seems somewhat similar to [JSON example](https://github.com/actix/examples/tree/master/json) so maybe we'd need some common example on extractors in future.
 But anyway it seems to be good as it is.

One thin I'd like you to do is to ad `README.md` file with short description how to use it and what it is.

form/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "form-example"
+version = "0.1.0"
+authors = ["Gorm Casper <gcasper@gmail.com>"]
+
+[dependencies]
+actix = "0.5"
+actix-web = "^0.6"
+serde = "1.0.69"
```
DoumanAsh: Didn't notice it at first glance...

Please use `1.0` for serde versions

casperin:
I definitely looked to the json example for how to structure it, although I didn't agree with using postman for testing. Maybe, with a merge, `curl` could be used as an example of requests.

DoumanAsh:
Ah this is understandable. I actually prefer simplicity so I see nothing wrong with your approach.

Last thing: can you please add check of your example in `.travis.yml`?

casperin:
Totally missed the travis file. It's done :)

DoumanAsh:
Yep, Thank you

PR #27, opened 2018-07-09
Middleware example

casperin:
I attempted to make them as minimal as possible, so I did not want to include other already implemented middlewares, for instance for checking the `req.identity()`.

I find that when I'm trying to learn a concept, I want as few distractions as possible.

### DIFF
--- a/middleware/src/main.rs
+++ b/middleware/src/main.rs
@@ -3,6 +3,9 @@ extern crate actix_web;
 
 use actix_web::{server, App};
 
+#[allow(dead_code)]
+mod redirect;
+#[allow(dead_code)]
 mod simple;
 
 fn main() {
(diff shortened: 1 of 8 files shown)


middleware/src/main.rs
```diff
@@ -0,0 +1,29 @@
...
+#[allow(dead_code)]
+mod simple;
+
+fn main() {
+    let sys = actix::System::new("middleware-example");
+
+    let _addr = server::new(|| {
+        App::new()
+            .middleware(simple::SayHi)
+            // .middleware(redirect::CheckLogin)
```
fafhrd91: could you enable this middleware or remove it?

casperin: I can, of course (which ever you prefer); but I wanted to isolate each middleware and not have them run on top of each other as that only makes it more difficult to understand what is happening.

I don't mind removing it either, but then the example would not have been helpful to myself (as a  very new to Rust dev) a week or two ago as I could not figure out how to conditionally halt a request.

fafhrd91:
thanks! we can modify example later

casperin:
Yeah please do. Go wild :)

And thanks for this awesome project!

PR #28, opened 2018-07-16
[WIP] update r2d2 example to actix 0.7

themasch:
I tried to update the r2d2 example to the latest actix version because I rand into an error I when trying to build something from it. 

Turns out, the same error occures in the example as well as soon as I updated it from actix 0.5 to 0.7 and fixed the type arguments of Addr<>. 

```


fafhrd91:
You need to switch actix-web to master

themasch:
Thanks for helping out @fafhrd91!
I guess its better than to wait for the next actix-web release so the examples are based on actual released version and not on some git master? 
Should I close the PR in the meantime?

PR #30, opened 2018-07-21
Use built-in auth middleware

sadika9:


fafhrd91:
this example is implemented this way specifically to show how impl complex middlewares

sadika9:
shouldn't it named specifically (e.g. complex-middleware) because it there is no example in the documentation for auth middleware?

I think it is better to have "cookie-auth" example using built-in middleware and have separate example for complex middleware implementation.

fafhrd91:
PR is welcome for separate example

PR #40, opened 2018-08-27
Fix json example Travis build failure

munckymagik:
When I raised #39 the Travis build was failing. This is a fix that makes it pass.

There were a couple of ways to fix this, I went for the "destructured arguments" approach. Another would have been something like:

```rust
   (cfg.0).0.limit(4096); // <- limit size of the payload
```

I felt the destructuring communicated more about what the closure argument looks like. Let me know if there's a better approach.


DoumanAsh:
Oh thanks

PR #41, opened 2018-08-28
fix missing heartbeat

botika:


websocket-tcp-chat/src/main.rs
```diff
@@ -183,6 +188,34 @@ impl StreamHandler<ws::Message, ws::ProtocolError> for WsChatSession {
     }
 }
 
+impl WsChatSession {
+    /// helper method that sends ping to client every second.
+    ///
+    /// also this method check heartbeats from client
+    fn hb(&self, ctx: &mut ws::WebsocketContext<Self, WsChatSessionState>) {
+        ctx.run_later(Duration::new(1, 0), |act, ctx| {
```
DoumanAsh: I think in this case you could use [run_interval](https://actix.rs/actix-web/actix_web/actix/actix/trait.AsyncContext.html#method.run_interval) to schedule periodical execution of this function

botika:
fix python client await `ws.pong()`

websocket-tcp-chat/src/main.rs
botika: Yes, I also change it in the implementation of session.rs

PR #42, opened 2018-08-28
fix missing websocket heartbeat and python client in websocket-tcp-chat

botika:


websocket-tcp-chat/src/main.rs
```diff
@@ -183,6 +188,34 @@ impl StreamHandler<ws::Message, ws::ProtocolError> for WsChatSession {
     }
 }
 
+impl WsChatSession {
+    /// helper method that sends ping to client every second.
+    ///
+    /// also this method check heartbeats from client
+    fn hb(&self, ctx: &mut ws::WebsocketContext<Self, WsChatSessionState>) {
+        ctx.run_later(Duration::new(1, 0), |act, ctx| {
```
DoumanAsh: Consider `run_interval`?

botika: Yes. I answered in the previous pull request, sorry. I change it also in the implementation of session.rs

botika:
No merging, please. yet I'm going to upload a fix to the client in python that does not send a messages

DoumanAsh:
Ok, ping me when I should review again.

botika:
Everything ready, you can check out pull request

websocket-tcp-chat/client.py
```diff
@@ -1,52 +1,60 @@
...
-    ws = yield from aiohttp.ClientSession().ws_connect(url, autoclose=False, autoping=False)
+    ws = await aiohttp.ClientSession().ws_connect(url, autoclose=False, autoping=False)
 
-    # input reader
     def stdin_callback():
         line = sys.stdin.buffer.readline().decode('utf-8')
         if not line:
             loop.stop()
         else:
-            ws.send_str(name + ': ' + line)
```
DoumanAsh: not sure I understand, why you needt o put `ws.send_str` into a separate queue?

I guess it is not related to heartbeats itself?

botika: In the current implementation, it does not send messages. Since it does not solve the routine in the callback. I fixed it.

botika: https://docs.python.org/3/library/asyncio-eventloop.html?highlight=add_reader#asyncio.AbstractEventLoop.add_reader Callback can not be solved in future and `ws.send_str` is solved in the future so it does not solve any of them. I discovered it because the client in python was disconnected by hb timeout.

DoumanAsh: is `send_str` must be awaitable?
I ghought it would be possible to use async callback here, but if not then ok

botika: I based myself on this post, https://stackoverflow.com/questions/29475007/python-asyncio-reader-callback-and-coroutine-communication . The asyncio documentation is little more than impossible

DoumanAsh: OK thanks for explanation. I'm not really familiar with python async story 😄

DoumanAsh:
thx

PR #44, opened 2018-09-16
WebSocket Chat example using actix-broker

chris-ricketts:


websocket-chat-broker/README.md
```diff
@@ -0,0 +1,29 @@
+# Websocket chat broker example
+
+This is a rewrite of the
```
DoumanAsh: Let's say it is different implementation of websocket chat example

websocket-chat-broker/src/main.rs
```diff
@@ -0,0 +1,156 @@
...
+
+        WsChatServer::from_registry()
+            .send(join_msg)
+            .map_err(|_| ())
+            .into_actor(self)
+            .map(|id, act, _| {
+                act.id = id;
+                act.room = room_name;
+            })
+            .wait(ctx);
```
DoumanAsh: It would be nice to avoid blocking in async code.
It is it possible to avoid relying on wait?

P.s. it is general comment for all futures that are transformed to ActorFuture

websocket-chat-broker/src/main.rs
```diff
@@ -0,0 +1,156 @@
...
+    type Result = (); 
+
+    fn handle(&mut self, msg: ChatMessage, ctx: &mut Self::Context) {
+        ctx.text(msg.0);
+    }
+}
+
+impl StreamHandler<ws::Message, ws::ProtocolError> for WsChatSession {
+    fn handle(&mut self, msg: ws::Message, ctx: &mut Self::Context) {
+        println!("WEBSOCKET MESSAGE: {:?}", msg);
```
DoumanAsh: debug println, maybe to use logging in this example?

websocket-chat-broker/src/main.rs
```diff
@@ -0,0 +1,156 @@
...
+        match msg {
+            ws::Message::Text(text) => {
+                let m = text.trim();
+                if m.starts_with('/') {
+                    let v: Vec<&str> = m.splitn(2, ' ').collect();
```
DoumanAsh: could you please have a better name for `v`?

Also I think collecting into Vec() is not required here, you could use just Split iterator.
If you feel like improving

websocket-chat-broker/src/main.rs
```diff
@@ -0,0 +1,156 @@
...
+    HttpServer::new(move || {
+        App::new()
+            .resource("/", |r| r.method(http::Method::GET).f(|_| {
+                HttpResponse::Found()
+                    .header("LOCATION", "/static/websocket.html")
+                    .finish()
+            }))
+            .resource("/ws/", |r| r.route().f(chat_route))
+            .handler("/static/", fs::StaticFiles::new("static/").unwrap())
+    }).bind("127.0.0.1:8080")
```
DoumanAsh: Just to be sure, could you please run rustfmt?

websocket-chat-broker/src/main.rs
```diff
@@ -0,0 +1,156 @@
...
+        }
+    }
+}
+
+fn main() {
+    let sys = actix::System::new("websocket-broker-example");
+
+    HttpServer::new(move || {
+        App::new()
+            .resource("/", |r| r.method(http::Method::GET).f(|_| {
```
DoumanAsh: The `/static` directory is already served below?
Maybe you could just rename it as `index.html` and server the whole static directory at `/`?

websocket-chat-broker/src/main.rs
```diff
@@ -0,0 +1,156 @@
...
+    room: String,
+    name: Option<String>,
+}
+
+impl WsChatSession {
+    fn join_room(&mut self, room_name: &str, ctx: &mut ws::WebsocketContext<Self>) {
+        let room_name = room_name.to_owned();
+        // First send a leave message for the current room
+        let leave_msg = LeaveRoom(self.room.clone(), self.id);
+        self.issue_sync(leave_msg, ctx);
```
DoumanAsh: Hmmm strangely enough I cannot find the `issue_sync` method, could you please point from where it is coming?

Chris-Ricketts:
Thanks @DoumanAsh - I'll look into those comments.

websocket-chat-broker/src/main.rs
Chris-Ricketts: Thanks for the hint. You were dead right - both those futures could be spawned instead of waited upon.

websocket-chat-broker/src/main.rs
Chris-Ricketts: This comes from having the `actix_broker::BrokerIssue` trait in scope. 
It implements the following functions for actors with an `AsyncContext`:
* `issue_async` - sends a message asynchronously to the message broker (uses `do_send` on the broker address from the system registry).
* `issue_sync` - sends a message synchronously to the message broker (uses `.wait(ctx)`).

After seeing your other comment about not blocking, it made me realise that maybe a third function needs to be added that returns an `ActorFuture`. This can then have computation chained on, such as the 'join room' future.

After seeing your comment about the

Chris-Ricketts:
@DoumanAsh  please see code changes and line comment replies.

websocket-chat-broker/README.md
Chris-Ricketts: Changed to "different implementation"

websocket-chat-broker/src/main.rs
Chris-Ricketts: Added `log` and `simple_logger` crates. Changed all `println!`'s.

websocket-chat-broker/src/main.rs
Chris-Ricketts: I just match on the iterator - much nicer to read :)

websocket-chat-broker/src/main.rs
Chris-Ricketts: Almost forgot but done!

websocket-chat-broker/src/main.rs
Chris-Ricketts: Sorted this out.

DoumanAsh:
Ok, thanks it looks great now.

One more small thing, could maybe it a bit info in readme why user would benefit from using `actix-broker` in comparison with other websocket example?
I guess it is not actually necessary, but I think it would be educational for these who look into examples.
What do you think?

Chris-Ricketts:
Thanks!

To be honest, I don't think `actix-broker` adds anything to this example apart from very slightly reducing the line count. There are only two actors, `WsChatServer` and `WsChatServer`, and they are aware of each others existence. `actix-broker` is more useful in larger actor systems where actors just want to subscribe to and broadcast messages, without really caring for a response. It eliminates cyclic dependencies in that case. @fafhrd91 asked me to rewrite the websocket-chat example in Gitter after I posted a link to the `actix-broker` repo.

PR #45, opened 2018-09-17
Fix "websocket" example's readme, and make binary names unique

sapir:


rustls/Cargo.toml
```diff
@@ -5,7 +5,7 @@ authors = ["Nikolay Kim <fafhrd91@gmail.com>"]
 workspace = "../"
 
 [[bin]]
-name = "server"
+name = "rustls-server"
```
DoumanAsh: Should you update `cargo run` in README?

tls/Cargo.toml
```diff
@@ -5,7 +5,7 @@ authors = ["Nikolay Kim <fafhrd91@gmail.com>"]
 workspace = "../"
 
 [[bin]]
-name = "server"
+name = "tls-server"
```
DoumanAsh: Should you update `cargo run` in README?

rustls/Cargo.toml
sapir: I think it's ok because there's only one binary in this directory

I get some "No such file or directory" panics when running it from the root directory, but then I get those even when running in its own directory with current master.

DoumanAsh:
Thank you

Issue #46, opened 2018-09-18
Incorrect "heart-beat" comments in websocket examples


sapir:
`websocket/src/client.rs` code says that if it doesn't do a heart-beat every 10 seconds, then server will disconnect it. But if I remove the heart-beat, then the client doesn't get disconnected.

The `websocket-chat/src/main.rs` server tracks heart-beats but doesn't do anything else with them. Again, a comment says "Client must send ping at least once per 10 seconds, otherwise we drop connection." But it doesn't really drop the connection.

Finally, in the `websocket-tcp-chat` example, both the server and the client do heart-beats, so if the client doesn't do its heart-beat, the server really does drop it.

PR #47, opened 2018-09-18
Remove unused heartbeats for websocket and websocket-chat.

sapir:
for #46


DoumanAsh:
Please take a look at error in build:

```
error[E0004]: non-exhaustive patterns: `Pong(_)` not covered
-->websocket-chat/src/main.rs:100:15
|
100|        match msg {
|^^^pattern `Pong(_)` not covered
```

On the side note, maybe instead of removing it would be better to make properly work?

sapir:
wouldn't it make the example overly complicated?

DoumanAsh:
I see, it might be true, but

I guess the question is then, it is necessary for full example of websocket?

PR #48, opened 2018-09-22
Implement websocket heartbeats

sapir:


websocket-chat/src/main.rs
```diff
@@ -173,11 +177,40 @@ impl StreamHandler<ws::Message, ws::ProtocolError> for WsChatSession {
...
         }
     }
 }
 
+impl WsChatSession {
+    /// helper method that sends ping to client every second.
+    ///
+    /// also this method checks heartbeats from client
+    fn hb(&self, ctx: &mut ws::WebsocketContext<Self, WsChatSessionState>) {
+        ctx.run_interval(Duration::new(1, 0), |act, ctx| {
```
DoumanAsh: Should maybe interval be a bit bigger? like 2 or 5 seconds?

websocket-tcp-chat/src/main.rs
```diff
@@ -112,8 +110,10 @@ impl StreamHandler<ws::Message, ws::ProtocolError> for WsChatSession {
     fn handle(&mut self, msg: ws::Message, ctx: &mut Self::Context) {
         println!("WEBSOCKET MESSAGE: {:?}", msg);
         match msg {
-            ws::Message::Ping(msg) => ctx.pong(&msg),
-            ws::Message::Pong(msg) => self.hb = Instant::now(),
+            ws::Message::Ping(msg) => {
+                self.hb = Instant::now();
```
DoumanAsh: shouldn't timing be updated on `Pong`?

sapir: this is to match the comment in the client code, that says that it must send a ping every 10 seconds, otherwise server disconnects the client.
timer could really be updated on any message. and maybe the client should also be checking for timeouts with server pings/pongs/whatever.

DoumanAsh: Hmmm now that you mention I guess it would make sense to update on every incoming message, since in the sense it means that remote peer is alive...

Though since both are supposed to send pings I guess we could leave it as it is.
Maybe add comment about it?

websocket-chat/src/main.rs
```diff
@@ -173,11 +177,40 @@ impl StreamHandler<ws::Message, ws::ProtocolError> for WsChatSession {
...
+            // check client heartbeats
+            if Instant::now().duration_since(act.hb) > Duration::new(10, 0) {
```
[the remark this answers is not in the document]
sapir: ok

websocket-chat/src/main.rs
sapir: ok

websocket-tcp-chat/src/main.rs
sapir: ok (if I understand correctly)

danno131313: Just made a pull request to revert this before actually seeing this one. Currently, there is no way to stay in a websocket connection through the browser for longer than 10 seconds because of this change. There is no way to ping the server through JS so the server must be the one to ping the client.

PR #49, opened 2018-09-28
make the code neater

Madao-3:
It shouldn't repeat.


DoumanAsh:
@Madao-3 Thanks, I guess there should be no problem with compilation of this example? I see Travis fails with something else

Issue #50, opened 2018-09-29
basics example: wrong index.html or the proper handler is missing


EasyToFindName:
The index.html in the "basics" example contains a chat form which trying to establish a websocket connection using the following url: "127.0.0.1:8080/ws" when I click the "Connect" button, but in the main.rs there is no resource named "/ws" and the ws module even isn't imported. So I get an "Error during WebSocket handshake: Unexpected response code: 404" and can't do anything useful with this chat window.

PR #51, opened 2018-09-29
Fix favicon's href

EasyToFindName:
Small fix to display favicon on the welome page tab. 

Also added the most likely fix of my issue: #50


DoumanAsh:
Oh... maybe it is copy-paste error with that example...
Pretty sure basic example is not supposed to contain websockets

PR #52, opened 2018-09-30
Fix pinging so client doesn't get dropped automatically after 10 seconds

danno131313:
The server would ping the client on a regular interval but never actually handled the pong, so the client disconnected after 10 seconds regardless.


DoumanAsh:
Thanks

PR #54, opened 2018-10-10
Encapsulate app state in Arc<Mutex<...>>

jannic:
As pythoneer argues in https://github.com/actix/actix-web/issues/531,
a realistic counter should count the absolute number of page loads,
and not per-thread counts.

Adding Arc<Mutex<...>> to the state example only increases the
complexity slightly, but is probably more helpful to a reader who just
wants to save some global state.

### DIFF
--- a/state/src/main.rs
+++ b/state/src/main.rs
@@ -15,21 +15,22 @@ extern crate actix;
 extern crate actix_web;
 extern crate env_logger;
 
-use std::cell::Cell;
+use std::sync::Arc;
+use std::sync::Mutex;
 
 use actix_web::{middleware, server, App, HttpRequest, HttpResponse};
 
 /// Application state
 struct AppState {
-    counter: Cell<usize>,
+    counter: Arc<Mutex<usize>>,
 }
 
 /// simple handle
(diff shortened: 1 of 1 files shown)


state/src/main.rs
```diff
@@ -15,21 +15,22 @@ extern crate actix;
...
-use std::cell::Cell;
+use std::sync::Arc;
+use std::sync::Mutex;
 
 use actix_web::{middleware, server, App, HttpRequest, HttpResponse};
 
 /// Application state
 struct AppState {
-    counter: Cell<usize>,
+    counter: Arc<Mutex<usize>>,
```
DoumanAsh: To avoid further confusion, let's make sure that people understand why state here should be `Arc`

Please add comment that in order for counter to be shared among all HTTP workers, we wrap it into `Arc<Mutext<T>>`

DoumanAsh:
Sorry about broken CI, the recent update of `tokio-rustls` broke some stuff.

state/src/main.rs
DoumanAsh: Ah you're right, I didn't check the whole file.
Then yes, leave it  as it is

DoumanAsh:
Thank you

PR #55, opened 2018-10-12
fix rustls example with updated rustls in actix-web

jannic:
actix-web updated to rustls 0.14. Update the example, as well, to avoid conflicting versions.
(to test this, one needs current master of actix-web, as well, to include commit 32145cf6 "fix after update tokio-rustls")


DoumanAsh:
Thank you

PR #56, opened 2018-10-12
replace deprecated feature "alpn" with "ssl"

jannic:
In actix-web's Cargo.toml, feature "alpn" is marked as deprecated,
with the suggestion to use "ssl", instead.


DoumanAsh:
Thank you

PR #59, opened 2018-11-08
Fix proxy to handle both chunked and non-chunked responses

vmalloc:


http-full-proxy/src/main.rs
```diff
@@ -11,7 +11,7 @@ use actix_web::{
     HttpResponse,
 };
 use clap::{value_t, Arg};
-use futures::Future;
+use futures::{future::ok, Future};
```
DoumanAsh: Do not import `ok` like that, I'd advise to use `future::ok` directly.

vmalloc: Got it. Fixed

http-full-proxy/src/main.rs
```diff
@@ -55,16 +55,29 @@ fn forward(
...
+fn construct_response(
+    resp: client::ClientResponse,
+) -> Box<dyn Future<Item = HttpResponse, Error = Error>> {
+    let mut client_resp = HttpResponse::build(resp.status());
+    for (header_name, header_value) in
+        resp.headers().iter().filter(|(h, _)| *h != "connection")
+    {
+        client_resp.header(header_name.clone(), header_value.clone());
+    }
+    //    Ok(client_resp.streaming(resp.payload()))
```
DoumanAsh: Oh yeah remove plz comment?

PR #62, opened 2018-11-15
remove unneeded RefCells

pjenvey:
interior mutability isn't needed within these actors as the handle
method takes &mut self

also fix compat. w/ rand 0.6


DoumanAsh:
Thanks

Issue #61, opened 2018-11-11
JSON post examples sending to other actors


Firstyear:
There are no example of using a JSON post where the request is sent to another actor, and the json deserialised correctly. The workflow in mind is:

```
index(request<AppState>) -> ... {
    req.payload()
        .from_err()
        .fold( ... bytes -> body)
        .and_then(body)  {
             let result_obj = serde_json::from_slice::<MyStruct>(&body);
             // send a custom error on Err(e) in result_obj
             // send the obj to another actor and then respond with json
             // IE request.state().actor.send().... 
        }
}
```

This is a reasonably complex example, incorporating many different components, and would be really important to have because there is no current example that demonstrates these components working together (json error handling and response, communicating to an actor, and handling the post). There are small elements of each through out the examples project, but without having a complete, tied together example it's really difficult to tie everything together.

scull7:
It's not a JSON post example, but I have made an example to show how to send information between actors #64

robjtede:
closing as stale; we're largely transitioning away from demonstrating actors in today's async landscape

PR #63, opened 2018-11-17
Add nonfunctional complex example to diesel

Firstyear:
Related to https://github.com/actix/examples/issues/61

This is a non-functional example of what I think is an important use case: Being able to post to an actor and then return a json response. It would be great to have this reviewed and corrected to functionally work.


diesel/src/main.rs
```diff
@@ -56,6 +59,59 @@ fn index(
...
 
+#[derive(Debug, Serialize, Deserialize)]
+struct MyUser {
+    name: String
+}
+
+const MAX_SIZE: usize = 262_144; // max payload size is 256k
+
+/// This handler manually load request payload and parse json object
+fn index_add(req: &HttpRequest<AppState>) -> Box<Future<Item = HttpResponse, Error = Error>> {
```
DoumanAsh: Should you use here `Json` extractor instead of manually parsing `MyUser`?

Firstyear: I don't see why I should be using a different extractor type, as this is a situation that should be workable. I have the requirement to add extra data into the error messages based on parsing issues, so I need to use the manual method to create those errors. I will also extend this example to have the extractor type, but I'd really appreciate advice on making this situation work. I've also got a similar example of the same with async that fails in it's own unique way. (I also seem to recall that the extractor type has a very similar set of issues and I can't get that working either ....)

Firstyear: As noted, extractor causes the following errors. I'll post the code to this PR now. 
```
error[E0277]: the trait bound `actix_web::HttpResponse: futures::Future` is not satisfied
   --> diesel/src/main.rs:125:10
    |
125 |         .and_then(|res| match res {
    |          ^^^^^^^^ the trait `futures::Future` is not implemented for `actix_web::HttpResponse`
    |
    = note: required because of the requirements on the impl of `futures::IntoFuture` for `actix_web::HttpResponse`

error[E0308]: mismatched types
   --> diesel/src/main.rs:119:5
    |
118 |   fn extract_item_limit((item, req): (Json<MyUser>, HttpRequest<AppState>)) -> HttpResponse {
    |                                                                                ------------ expected `actix_web::HttpResponse` because of return type
119 | /     req.state()
120 | |         .db
121 | |         .send(CreateUser {
122 | |             name: item.name,
...   |
127 | |             Err(_) => HttpResponse::InternalServerError().into(),
128 | |         })
    | |__________^ expected struct `actix_web::HttpResponse`, found struct `futures::AndThen`
    |
    = note: expected type `actix_web::HttpResponse`
               found type `futures::AndThen<futures::future::FromErr<actix::prelude::Request<db::DbExecutor, db::CreateUser>, _>, actix_web::HttpResponse, [closure@diesel/src/main.rs:125:19: 128:10]>`

error[E0282]: type annotations needed
   --> diesel/src/main.rs:153:55
    |
153 |                     .with_config(extract_item_limit, |((cfg, _),)| {
    |                                                       ^^^^^^^^^^^ consider giving this closure parameter a type
    |
    = note: type must be known at this point

error: aborting due to 4 previous errors

```

diesel/src/main.rs
```diff
@@ -112,6 +113,19 @@ fn index_add(req: &HttpRequest<AppState>) -> Box<Future<Item = HttpResponse, Err
         .responder()
 }
 
+fn extract_item_limit((item, req): (Json<MyUser>, HttpRequest<AppState>)) -> HttpResponse {
```
DoumanAsh: You return future in this function, but function signature is `HttpResponse`

diesel/src/main.rs
DoumanAsh: I suspect there might be ambiguous type in `with_config` because you failed to compile `extract_item_limit`

diesel/src/main.rs
```diff
@@ -56,6 +59,73 @@ fn index(
...
+struct MyUser {
+    name: String
+}
+
+const MAX_SIZE: usize = 262_144; // max payload size is 256k
+
+/// This handler manually load request payload and parse json object
+fn index_add(req: &HttpRequest<AppState>) -> Box<Future<Item = HttpResponse, Error = Error>> {
+    // HttpRequest::payload() is stream of Bytes objects
+    req.payload()
```
DoumanAsh: You probably want here not request, but state extractor https://actix.rs/actix-web/actix_web/struct.State.html

diesel/src/main.rs
Firstyear: Hey there, that seems to be separate from the issue. I have found no way to make cfg work with the with_config type and a future response.

diesel/src/main.rs
Firstyear: There seems to be no way to mix a config and state extractor.

I also want to say, I really appreciate your feedback here, and I know you are trying to "advise" to a simpler way, but there a good reasons that I am asking for the examples I am. I want to be able to respond with complex error messages about failures in the parsing of the body (extractors block this), I want to eventually be able to decode types other than json (CBOR etc). I want to have DoS protections (message size limits and more). I also want to be able to standardise on a single json parser (serde). 

So I really appreciate your advice here, but I would really like your help to fix the examples I have here: both the manual parsing example and the extractor async example. I'm sure that other people will run into these problems in the future, and so getting good, advanced use cases that are production ready as examples will really help adoption of the framework,

Thanks,

DoumanAsh:
Ehhh ok, I'll check it out today at evening

Firstyear:
Thank you, I do really appreciate the help and the lesson.

DoumanAsh:
Take a look at my commit, I fixed it

diesel/src/main.rs
```diff
@@ -172,8 +170,8 @@ fn main() {
                     //  below).
                     //
                     // This does not work
-                    .with_config(add2, |cfg| {
-                        cfg.0.limit(4096); // <- limit size of the payload
+                    .with_async_config(add2, |(json_cfg, )| {
```
DoumanAsh: For methods that return future as response, you need `with_async*`, not plain `with`

diesel/src/main.rs
```diff
@@ -88,46 +88,44 @@ fn index_add(req: &HttpRequest<AppState>) -> Box<Future<Item = HttpResponse, Err
...
-            Ok(user) => Ok(HttpResponse::Ok().json(user)),
-            Err(_) => Ok(HttpResponse::InternalServerError().into()),
-        }))
+fn add2((item, state): (Json<MyUser>, State<AppState>)) -> impl Future<Item = HttpResponse, Error = Error> {
+    state.db
+         .send(CreateUser {
+             // Is it possible to do with without cloning? Apparently this is "borrowed".
+             // A risk is that when this is sent as a message, the lifetimes will
+             // be very complex to handle, and may not be reasonable ...
+             name: item.name.clone(),
```
DoumanAsh: It is borrowed due to `Deref`, use `Json::into_innter` to get `MyUser` struct and then move out of it

Firstyear:
Hey there, that looks really good. I'm reviewing it to learn more about what I did wrong.

I've uncommented the index_add function and it's not compiling though:

```
error[E0631]: type mismatch in function arguments
   --> diesel/src/main.rs:161:64
    |
71  | fn index_add((req, state): (HttpRequest<AppState>, State<AppState>)) -> impl Future<Item = HttpResponse, Error = Error> {
    | ----------------------------------------------------------------------------------------------------------------------- found signature of `fn((actix_web::HttpRequest<AppState>, actix_web::State<AppState>)) -> _`
...
161 |             .resource("/add", |r| r.method(http::Method::POST).f(index_add))
    |                                                                ^ expected signature of `for<'r> fn(&'r actix_web::HttpRequest<AppState>) -> _`


```

I tried some other "types" like .f(index_add), .a(index_add), but I'm not clear on what I'm missing. Maybe the state extractor or similar?

Once that's fixed up I'll clean up the patches then would you be happy to merge this example?

DoumanAsh:
Look at `with_async` methods

Firstyear:
Fantastic! That indeed did the trick. Would you like the code cleaned up in anyway as an example for merging? Rebase squash, or anything else.

DoumanAsh:
Yes please prepare PR for final review

Firstyear:
I've squashed the commits down, reworked the commit message and checked the patch seems reasonable. Is there anything else that I need to do before this is accepted?

Firstyear:
No problem, I'll fix all of this up. Thanks!

Firstyear:
Thanks @DoumanAsh I've rebased and pushed, and that should fix the comments you raised.

diesel/src/main.rs
```diff
@@ -56,6 +60,73 @@ fn index(
...
+                               });
+
+                Box::new(res)
+                }
+                Err(_) => Box::new(future::err(error::ErrorBadRequest("Json Decode Failed")))
+            }
+        })
+}
+
+fn add2((item, state): (Json<MyUser>, State<AppState>)) -> impl Future<Item = HttpResponse, Error = Error> {
```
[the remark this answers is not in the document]
DoumanAsh: Comment is not addressed

DoumanAsh:
Remove `test.db` file from repo

Firstyear:
Removed test.db, and rebased to master.

diesel/Cargo.toml
```diff
@@ -19,3 +20,5 @@ serde_derive = "1.0"
 diesel = { version = "^1.1.0", features = ["sqlite", "r2d2"] }
 r2d2 = "0.8"
 dotenv = "0.10"
+
+json = "*"
```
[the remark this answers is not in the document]
Firstyear: This dependency is used :)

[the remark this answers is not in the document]
DoumanAsh: Could you please try to point out where? I tried to grep in your changes but cannot see aside from `extern crate json`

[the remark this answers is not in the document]
Firstyear: It looks like human error on my part, and not saving the change where I removed the extern crate. Runnig again, you are correct that it's not needed. Pushing an update now.

DoumanAsh:
Thanks

PR #68, opened 2018-12-15
Added a tutorial link in the readme

mygnu:
I thought it might be helpful to users if they are looking for a detailed tutorial. If you don't want it that's ok too :)


mygnu:
I'm not sure if that was the right thing to do, I'm going to create another branch just for that change

PR #69, opened 2018-12-15
Readme updated with a tutorial link

mygnu:
I thought it might be helpful to users if they are looking for a detailed tutorial. If you don't want it that's ok too :)


mygnu:
Thank you

PR #70, opened 2018-12-28
Introduce actix-diesel and use in the diesel example

mehcode:
Had a lot of fun changing this example. I tried to keep it as unchanged as possible. Check out https://github.com/mehcode/actix-diesel/tree/master/example for a more involved example of how nice the ergonomics can get with `async fn` with this.

`actix-diesel` essentially just uses an Actor that ships your closure around and executes it in a synchronous thread pool with a bunch of magic type signatures to make it work. The other improvement is in the state management. It defers creation of the arbiter until first use so it can piggy back on the System created by `server::new ... .run()` and not require users to change their Actix Web setup.


DoumanAsh:
To be honest I think it would be better to leave original actix-diesel example without third party library.

Users might need such examples as it is general approach to blocking code, so I'd suggest to make it as separate example, not as replacement

diesel/src/main.rs
```diff
@@ -1,84 +1,69 @@
...
 ) -> FutureResponse<HttpResponse> {
-    // send async `CreateUser` message to a `DbExecutor`
     state
         .db
-        .send(CreateUser {
-            name: name.into_inner(),
-        })
-        .from_err()
-        .and_then(|res| match res {
+        .get(move |conn| db::create_user(conn, name.into_inner()))
```
DoumanAsh: So basically it is actor that allows you to get connection to work with it?

PR #78, opened 2019-02-18
Add example template_yarte

botika:
https://github.com/rust-iendo/yarte

#### Benchs
https://travis-ci.org/rust-iendo/yarte/jobs/495735070#L612

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ members = [
   "static_index",
   "template_askama",
   "template_tera",
+  "template_yarte",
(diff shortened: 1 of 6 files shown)


DoumanAsh:
I don't want to discourage you seeing as you want to introduce your template library, but looking at it, yarte is quite WIP itself, are you sure it wouldn't be better to stabilize it a bit before pushing such example?

botika:
Ok, we wait for version 0.1.0 in a few weeks.

PR #79, opened 2019-02-20
Add udp echo server example

apatrushev:


udp-echo/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "udp-echo"
+version = "0.1.0"
+authors = ["Anton Patrushev <apatrushev@gmail.com>"]
+edition = "2018"
+
+[dependencies]
+actix = "0.7"
+tokio = "*"
```
DoumanAsh: use `0.x` format for all deps, otherwise in future they'll break

udp-echo/src/main.rs
```diff
@@ -0,0 +1,47 @@
+extern crate actix;
```
DoumanAsh: with 2018 edition mode, you're not required to `extern crate` your deps

udp-echo/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+impl Actor for UdpActor {
+    type Context = Context<Self>;
+}
+
+#[derive(Message)]
+struct UdpPacket(BytesMut, SocketAddr);
+impl StreamHandler<UdpPacket, std::io::Error> for UdpActor {
+    fn handle(&mut self, msg: UdpPacket, _: &mut Context<Self>) {
+        println!("Received: ({:?}, {:?})", msg.0, msg.1);
+        (&mut self.sink).send((msg.0.into(), msg.1)).wait().unwrap();
```
DoumanAsh: If I'm not mistaken `Stream::send` should consume self? Does it compile?

apatrushev: It is `Sink::send` not a `Stream::send`.
https://docs.rs/futures/0.1.25/futures/sink/trait.Sink.html#method.send

DoumanAsh: Yeah ok, it doesn't change the fact that it takes `self` rather than `&mut self`

apatrushev: Interesting. But this code works somehow. :(

apatrushev: Ok, the travis checks passed showing that this code compiles. Should I fix it if it works? I do not like the situation when I do not understand why it works and will be happy to find explanation, but I definitely need a help.

DoumanAsh: If it compiles then everything is ok I guess. which is surprising but fine too

svenstaro:
Can you rebase this?

apatrushev:
> Can you rebase this?

Do you plan to merge it? I will do rebase in that case.

svenstaro:
I think it's a useful example and I'd like to merge it, yes.

apatrushev:
> I think it's a useful example and I'd like to merge it, yes.

Done.

svenstaro:
Thanks!

PR #80, opened 2019-02-21
Add file thread example

apatrushev:


file-thread/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+name = "main"
+path = "src/main.rs"
+
+[[bin]]
+name = "tap"
+path = "src/tap.rs"
+
+[dependencies]
+actix = "0.7"
+bytes = "*"
```
DoumanAsh: `0.1`

file-thread/src/main.rs
```diff
@@ -0,0 +1,71 @@
+use std::{env, fs, io::Read, process, thread, time};
+
+extern crate actix;
```
DoumanAsh: extern crate is not necessary

file-thread/src/main.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+fn main() {
+    let sys = actix::System::new("file-reader");
+
+    let path = env::args().skip(1).next().expect("please provide filepath");
+    let mut file = fs::File::open(path).expect("open failed");
+    let addr = FileProcessor {}.start();
+
+    thread::spawn(move || {
```
DoumanAsh: It seems fine for one time reading of file... but it doesn't really seem all that useful in practice.
I'd consider the use case where user might need to read file or different files from time to time. So maybe you could consider example of system that could provide user with convenient API to request file content?

apatrushev: I wrote about real use cases in README. Reading pipe or device file is a good target for this solution.

file-thread/Cargo.toml
apatrushev: Mine idea was that actix itself depends on bytes crate, so this code can rely on its dependency restriction. If it is a bad practice I will change it.

DoumanAsh: if you use `*` while actix uses `0.1` you gonna use two different crates when `0.2` appears

file-thread/src/main.rs
DoumanAsh: I doubt most people would read README, what they would be interested is real examples that they can copy-paste.

The way, the current example is shown, is pretty much one time read of file and that's all.
There is nothing in particular related to actix per se, having for example actor that would start reading file when receiving message would be more complex and useful, I think?

On side note, @Dowwie review please

apatrushev: Ok. How to read a pipe or device where the data coming over time? As I mentioned in gitter it is not a trivial task in actix. That was the starting point of my investigation. fectl pipe solution besides its complexity is not working with latest tokio (at least I did not find a way to fix it). I will not fight for this PR. It is your choice to adopt it or not. I just closed my personal obligations to share my experience and created working example.

robjtede:
closing since blocking ops are trivial in modern actix

PR #87, opened 2019-03-11
Add test to hello-world

allevo:
` .run(&index)?;` is wrong. I don't know how to translate it for `1.x` version.

Could someone help me?

### DIFF
--- a/hello-world/Cargo.toml
+++ b/hello-world/Cargo.toml
@@ -8,3 +8,6 @@ edition = "2018"
 [dependencies]
 env_logger = "0.6"
 actix-web = { git="https://github.com/actix/actix-web.git", branch = "1.0" }
+
+[dev-dependencies]
+bytes = "0.4"
--- a/hello-world/src/main.rs
+++ b/hello-world/src/main.rs
@@ -25,27 +25,38 @@ mod tests {
     use super::*;
 
     use actix_web::dev::Body::Bytes;
-    use actix_web::dev::ResponseBody::{Other};
-    use actix_web::HttpResponse;
+    use actix_web::dev::ResponseBody::{Body};
     use actix_web::Error;
     use actix_web::{http, test};
+    use actix_web::http::{Method};
 
      #[test]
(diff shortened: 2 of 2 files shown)


fafhrd91:
there are two options:

1. test full pipeline, https://github.com/actix/actix-web/blob/1.0/src/route.rs#L435
2. test only handler function, in that case you can use `actix_web::test::block_on(index(req))`

allevo:
I've used `init_service` but I'm able to test the response body. Which is the best way to do that?

Many thanks for the support.

allevo:
Another question: I would like to refactor the code in order to extract the `App::new()...` invocation from the main in order to call it into the test. I don't know which type should be returned by the `create_app` function:

```rust
fn create_app () -> ???? {
    App::new()
        .middleware(middleware::Logger::default())
        .service(web::resource("/index.html").to(|| "Hello world!"))
        .service(web::resource("/").to(index));
}
```

In this way the main could be written with
```
    HttpServer::new(|| create_app())
        .bind("127.0.0.1:8080")?
        .run()
```

Thanks

fafhrd91:
`create_app()` is problematic. you would need to write all types that App uses, even with impl Trait
it is not easy. trait aliases will help but it is nightly atm

fafhrd91:
regarding body, we need to add `Stream` impl for `MessageBody` types. 

would you like to create PR for actix-http?

allevo:
> create_app() is problematic. you would need to write all types that App uses, even with impl Trait
> it is not easy. trait aliases will help but it is nightly atm

Ok, but this is a pity.

> regarding body, we need to add Stream impl for MessageBody types.
> would you like to create PR for actix-http?

If I understood correctly is no possible to read the response body at all writing a code like:
```rust
let response = test::call_success(&mut srv, req);
let response_body = response.body().into_bytes();
```
Right?

If I have to send a PR to `actix-http`, could you give me some information about how and where to implement the missing code?

Many thanks

PR #88, opened 2019-03-12
Add template yarte

botika:
I would like to emphasize the two functionalities: 
 - Partials if you throw `cargo run` you will see the result by stdout. 
 - Hierarchical name resolution which allows to resolve and create names in different scopes with the minimum code. Resolve within any context, including tuples with `_n` for the n-th element.

The result is modular templates by composition without cost.


botika:
Close in favor of actix_web `1.0`

PR #97, opened 2019-03-28
Fix actix_redis to master

botika:
Confirm me if it is the right way and I finish the ones that remain.


fafhrd91:
actix-redis must be updated to actix-0.8

botika:
Ok, now I get actix_redis. @fafhrd91 But the `.data(AppState)` is how you thought?

PR #102, opened 2019-04-01
Fixes parentheses in template_askama main.rs

xaxoxa:
Switches the position of a rogue parenthesis so that the index function is actually called.


fafhrd91:
Thanks!

PR #107, opened 2019-04-08
Revert to alpha3 multipart example to avoid panic

xaxoxa:
The alpha4 multipart example seems to cause a panic (as mentioned in #736). So this reverts to the alpha3 example, tweaked for the actix-multipart crate.

https://github.com/actix/actix-web/issues/736


fafhrd91:
Thanks

PR #117, opened 2019-05-05
fix return home link bug

PrivateRookie:
As titled

### DIFF
--- a/basics/static/404.html
+++ b/basics/static/404.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html><html><head><title>actix - basics</title>
     <link rel="shortcut icon" type="image/x-icon" href="/favicon.ico" /></head>
   <body>
-    <a href="static/index.html">back to home</a>
+    <a href="../static/welcome.html">back to home</a>
     <h1>404</h1>
   </body>
 </html>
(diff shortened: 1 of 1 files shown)


basics/static/404.html
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html><html><head><title>actix - basics</title>
     <link rel="shortcut icon" type="image/x-icon" href="/favicon.ico" /></head>
   <body>
-    <a href="static/index.html">back to home</a>
+    <a href="../static/welcome.html">back to home</a>
```
DoumanAsh: `/static/welcome.html`?

PrivateRookie: > `/static/welcome.html`?

`/static/welcome.html` will not work, if you click "back to home", it will jump to http://localhost:8080/static/static/welcom.html, rather than http://localhost:8080/static/welcome.html

DoumanAsh: Eh? Is that so?

Because what you describe is behavior of `static/welcome.html` which is relative link to the current page
While `/static/welcome.html` should relative to site's root

PrivateRookie: sorry for my typo, both `/static/welcome.html` and `../static/welcome.html` work. So you prefer to use `/static/welcome.html` ？

DoumanAsh: Yeah, it is more reliable this way as it is relative to site's root.

DoumanAsh:
Thanks

PR #118, opened 2019-05-08
Arbiter::exec to start TcpServer will not work, simply call it

### LINKED ISSUE #115: websocket-tcp-chat does not open TCP listener
Not sure it's a bug in the example, in Actix or on my Mac but I cannot get the example to listen on TCP port 12345. `session::TcpServer::new` is never executed for some reason.

With the following change it works:
```
-    Arbiter::new().exec(move || {
-        session::TcpServer::new("127.0.0.1:12345", srv);
-        Ok::<_, ()>(())
-    });
+    session::TcpServer::new("127.0.0.1:12345", srv);
```

therealprof:
Fixes #115

Signed-off-by: Daniel Egger <daniel@eggers-club.de>

### DIFF
--- a/websocket-tcp-chat/src/main.rs
+++ b/websocket-tcp-chat/src/main.rs
@@ -1,5 +1,3 @@
-use serde::{Deserialize, Serialize};
-
 use std::time::{Duration, Instant};
 
 use actix::*;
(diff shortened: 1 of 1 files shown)


therealprof:
@fafhrd91 Any change of getting this simple fix for a broken example applied?

therealprof:
@JohnTitor Mind having a look?

JohnTitor:
Does this tcp listener work correctly now? It seems not to work well on my local. Could you make sure it?

therealprof:
Yes, it works for me after the PR.

therealprof:
The client seems to have an unrelated problem to me but at least the the server listens and receives joins and disconnects from the client.

JohnTitor:
Ah, it seems so, we should fix it later (and I should have noticed that...).

JohnTitor:
Thanks!

PR #127, opened 2019-05-24
Fix bug in database script

senden9:
Fix error that occurs if script is used like in the async_db readme.

Error before this change:
``` bash
$ LANG=C bash db/setup_db.sh
db/setup_db.sh: line 2: db.sql: No such file or directory
Error: cannot open "nyc_centralpark_weather.csv
```


fafhrd91:
thanks!

PR #129, opened 2019-05-25
Fix bug in state type

blackjune:
Fix bug in state type 
```rust
let counter = Arc::new(Mutex::new(0));
```
may not be usize

Error before this change:
```
App data is not configured, to configure use App::data()
```


fafhrd91:
Thanks!

Issue #123, opened 2019-05-17
'websocket' example appears to be out of date for client.rs


mcarlin:
I might be missing something but the client.rs doesn't appear to up to date with the latest APIs.  If I uncomment the client binary in the `Cargo.toml`, I see the following errors:
```
cargo build
   Compiling websocket v0.1.0 (/Users/mike/code/examples/websocket)
error[E0432]: unresolved imports `actix_web::client::ClientWriter`, `actix_web::client::Message`, `actix_web::client::ProtocolError`
 --> websocket/src/client.rs:6:33
  |
6 | use actix_web::client::{Client, ClientWriter, Message, ProtocolError};
  |                                 ^^^^^^^^^^^^  ^^^^^^^  ^^^^^^^^^^^^^ no `ProtocolError` in `client`
  |                                 |             |
  |                                 |             no `Message` in `client`
  |                                 no `ClientWriter` in `client`
  |                                 help: a similar name exists in the module: `ClientBuilder`

warning: unused import: `Message`
 --> websocket/src/client.rs:6:47
  |
6 | use actix_web::client::{Client, ClientWriter, Message, ProtocolError};
  |                                               ^^^^^^^
  |
  = note: #[warn(unused_imports)] on by default

error[E0191]: the value of the associated type `Result` (from the trait `_impl_act_clientcommand::actix::Message`) must be specified
  --> websocket/src/client.rs:87:20
   |
87 | impl StreamHandler<Message, ProtocolError> for ChatClient {
   |                    ^^^^^^^ associated type `Result` must be specified

error[E0191]: the value of the associated type `Result` (from the trait `_impl_act_clientcommand::actix::Message`) must be specified
  --> websocket/src/client.rs:88:31
   |
88 |     fn handle(&mut self, msg: Message, ctx: &mut Context<Self>) {
   |                               ^^^^^^^ associated type `Result` must be specified

error: aborting due to 3 previous errors

Some errors occurred: E0191, E0432.
For more information about an error, try `rustc --explain E0191`.
error: Could not compile `websocket`.

```

jesskfullwood:
As I mentioned when this was brought up in the [main repo](https://github.com/actix/actix-web/issues/829), I have tried to fix this, and I think doing so uncovers some weaknesses in the API. In particular:

[...]

PR #143, opened 2019-06-16
Update websocket client example to actix-web 1.0

guoli-lyu:
In this PR, a working example for websocket client is given.

The new API split Framed object into sink and stream to obtain writer and reader. And I found a helper struct `SinkWrite` in `actix::io`, which can hold sink object and provide an easy way to send message through Sink, as `Sink::send(self, item:Self::Item)` consumes Sink object.

Related to #123 and this issue in [actix-web](https://github.com/actix/actix-web/issues/829).

## Dependencies of server and client are quite different

After a few days of trial and error, I have figure out a way to write websocket client with actix-web 1.0, or rather, awc and actix-codec. So maybe it's better to create a new example folder, as the dependencies in server and client are quite different.

Minimal dependencies for server:
```toml
[dependencies]
actix = "0.8.2"
actix-web = "1.0.0"
actix-web-actors = "1.0.0"
actix-files = "0.1.1"
futures = "0.1"
bytes = "0.4"
```

Minimal dependencies for client:
```toml
[dependencies]
actix  = "0.8.2"
awc = "0.2.1"
actix-rt = "0.2.2"
actix-codec = "0.1.2"
futures = "0.1"
```

## Maybe simplify ChatClient struct

The previous ChatClient is `struct ChatClient(ClientWriter)`, and the new ChatClient seems a bit complicated:

```rust
struct ChatClient<T>(SinkWrite<SplitSink<Framed<T, Codec>>>)
where
    T: AsyncRead + AsyncWrite
```

Maybe it can be simply to `struct ChatClient<S: Sink>(SinkWrite<S>)`. I have tried but the compiler complains that only `Sink::Item` are acceptable when sending message.

### DIFF
--- a/websocket/Cargo.toml
+++ b/websocket/Cargo.toml
@@ -9,15 +9,17 @@ workspace = ".."
 name = "websocket-server"
 path = "src/main.rs"
 
-#[[bin]]
-#name = "websocket-client"
-#path = "src/client.rs"
+[[bin]]
+name = "websocket-client"
+path = "src/client.rs"
 
 [dependencies]
 actix = "0.8.2"
+actix-codec = "0.1.2"
 actix-web = "1.0.0"
 actix-web-actors = "1.0.0"
 actix-files = "0.1.1"
+awc = "0.2.1"
 env_logger = "0.6"
 futures = "0.1"
 bytes = "0.4"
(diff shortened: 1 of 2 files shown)


fafhrd91:
thanks!

PR #147, opened 2019-06-19
Avoid double Arc inside template_handlebars

linouxis9:
It isn't necessary to encapsulate the `Handlebars` struct inside an atomic reference-counting pointer, as `actix_web::web::Data` already uses an `Arc` internally (as specified in https://docs.rs/actix-web/1.0.2/actix_web/web/struct.Data.html).

Thanks ! :)


fafhrd91:
Thanks

PR #165, opened 2019-08-19
Remove old http-proxy, use the full version as the official http-proxy example

vmalloc:


http-proxy/Cargo.toml
```diff
@@ -1,21 +1,15 @@
 [package]
-name = "http-proxy"
+name = "http-full-proxy"
```
JohnTitor: IMO: The package, workspace, and directory name should have the same one. What about?

svenstaro: Sorry, what do you mean?

svenstaro: Oh I see, yes. It should be named `http-full-proxy`.

vmalloc: Oops, missed that. Fixed

http-proxy/Cargo.toml
```diff
@@ -1,21 +1,15 @@
 [package]
-name = "http-proxy"
+name = "http-full"
```
JohnTitor: Ah, the name is still invalid.
```suggestion
name = "http-proxy"
```

vmalloc: Oops, I feel stupid now. Fixed!

vmalloc:
I restored the SSL feature dependency because it should be able to use https, of course...

http-proxy/Cargo.toml
```diff
@@ -1,21 +1,14 @@
 [package]
 name = "http-proxy"
 version = "0.1.0"
-authors = ["Nikolay Kim <fafhrd91@gmail.com>"]
-edition = "2018"
+authors = ["Rotem Yaari"]
```
leizzer: is it needed to change the author? In any case why not listing both? (real question, don't know how the community handle crate authors)

Dowwie: I've authored several examples and haven't bothered adding myself to the list.  Nikolay has maintained the repo after initial submissions are made.   An original author plus Nikolay seems fair.

Thoughts?

vmalloc: Sure, I will add myself with Nikolai going first of course. The reason for this is that I just moved the full http-proxy example that I wrote from scratch over to this new one...

PR #167, opened 2019-09-01
Prevent jobs from being terminated on Travis

JohnTitor:
Fixes some cache errors on Travis
ref. https://levans.fr/rust_travis_cache.html

Other opinions and advice are welcome.

**EDIT**: and it seems clippy is not working so I try to fix


JohnTitor:
I'm going to merge this to make other PRs happy

PR #173, opened 2019-09-13
add custom middleware sound like hacker.

megumish:
I wrote a bit interesting example about middleware.

This is more complex than in /middleware, so I put it in /middleware-complex.

Here is my original repo
https://github.com/megumish/actix-web-original-middleware-example

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ members = [
   "jsonrpc",
   "juniper",
   "middleware",
+  "middleware-complex",
   "multipart",
   "protobuf",
   "r2d2",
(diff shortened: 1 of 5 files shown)


Dowwie:
What does this toy example demonstrate that isn't already demonstrated by the other middleware example?

megumish:
Umm ...
I maked this pr because  I got that I'm told to suggest to try.
When I will have gotten it that this pr is different  other, I will suggest that.

This is my first challenge to contribute OSS, so I'm sorry but I want you to teach if I'm doing do strange.

Dowwie:
@megumish Please, don't take this as discouragement from contributing.  In this PR, you published a simple example that covers no more functionality than that which was covered by the /middleware example.  Further, it uses playful, misleading terminology ("hacking") that is never used in a professional context as you have here.  People may still find your separate github repo useful.

megumish:
@Dowwie I see. When I come up with example more useful and cover than others, I try to submit pr again. Thanks for your review :)

PR #176, opened 2019-09-18
Add middleware to log the response body

svenstaro:
This adds an example showing off how to log out the response body inside of a middleware.
I'd like to get this thoroughly checked as I'm using a `wait()` in order to resolve the bytes so that I can feed them back into the response after moving the body out. @fafhrd91 hinted that this may cause problems but I'm not seeing any problems and this code works and produces expected results. 

@fafhrd91 also hinted that I might have to use a custom response as in [here](https://github.com/actix/actix-web/blob/master/src/middleware/logger.rs#L218) but I think that's really inconvenient and I'm not quite sure how to recover the body read out that way and put it back into the original middleware so I'd like find a less complicated solution if possible.


middleware/src/read_response_body.rs
```diff
@@ -0,0 +1,111 @@
...
+    type Response = ServiceResponse<BodyLogger<B>>;
+    type Error = Error;
+    type Future = WrapperStream<S, B>;
+
+    fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+        self.service.poll_ready()
+    }
+
+    fn call(&mut self, req: ServiceRequest) -> Self::Future {
+        WrapperStream {
```
fafhrd91: you can use box instead

svenstaro: Sorry, use `Box` instead of what exactly? I surely can't replace the whole `WrapperStream` with a `Box`?

svenstaro: I merged it but I'd still like you to explain your suggestion :D

Issue #32, opened 2018-07-26
Adding SSE examples using HttpContext and Channel


fafhrd91:

greenteamsite:
It would be very useful for me

Songtronix:
Does 1.0 still have HttpContext? Can't find it.

arve0:
I made [this example](https://github.com/arve0/actix-sse/blob/master/src/main.rs), which uses a [mutex lock](https://doc.rust-lang.org/std/sync/struct.Mutex.html) along with a [tokio bytes channel](https://docs.rs/tokio/0.1.22/tokio/sync/mpsc/fn.channel.html?search=). Not sure how it scales, will `Sender::try_send` block? 

Also, what would be a better way to detect stale clients? Implement an actor in `new_client` (the event source handler), which keeps track the connection state?

stellanhaglund:
Is that the best way to detect client disconnects, couldn't it be handle by detecting the tcp disonnect?

PR #177, opened 2019-09-19
server-sent events

arve0:
Issue #32, without `HttpContext` though.

~~Lacks mechanism to remove clients / connections when not receiving broadcast messages, which may starve file descriptors and hinder new HTTP connections.~~

Edit: Updated to clean stale clients every 10 seconds.


fafhrd91:
good example!

Issue #73, opened 2019-01-03
Add example for unit tests


leplatrem:
It would extremely useful to see a well-structured test suite, where requests are forged and sent to a test server in order to assert responses content.

Does anyone know an existing project that does that well out there? I would be happy to contribute a small example if there's a model I can be inspired by ;)

augustocdias:
I have one of a project I'm working for my company. 
https://github.com/augustocdias/rustbier/
You can check the `tests` directory
In my perspective, the problem is the external dependencies... (which the actual app will be one as well).
It might not be the best example that you're expecting because my API basically returns images and not HTML/JSON responses, but you can have an idea on how it works.
Bear in mind that I haven't adapted the code to be in github yet, because it was originally meant to be only in our internal gitlab. To make it work, you'll have to adapt some files. Basically you'll have to change the docker tag names in the `Makefile`, the `Dockerfile` to point to the base image you just built instead the one pointing to our internal registry (I made 2 images, because opencv and imagemagick take too long to compile), the `docker-compose` to point to the official `minio` (our have a embedded client packaged) image and the `setup.sh` to add the images to the `minio` container.

leizzer:
You can use any 3rd party tool you like or create a client that assert the response. (some of the [examples](https://github.com/actix/examples) have python clients to test the examples)

robjtede:
many examples have tests now: https://github.com/actix/examples/search?q=test

PR #179, opened 2019-09-29
Provide tests for the form example

pfrenssen:
It would be helpful to have tests for the various examples (ref. #73).

This PR adds unit tests and integration tests for all the POST request handlers in the form example.

I have factored out the app configuration to a separate function so it can be reused in the integration tests. The version bump is needed because the integration tests are using `test::TestRequest::set_form()` method which was added recently in https://github.com/actix/actix-web/pull/1058

This is my first contribution to actix-web so if there are certain things that can be done in an easier or better way I would be happy to improve the PR.

### DIFF
--- a/form/Cargo.toml
+++ b/form/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2018"
 workspace = ".."
 
 [dependencies]
-actix-web = "1.0.0"
+actix-web = "1.0.8"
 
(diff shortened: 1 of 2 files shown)


JohnTitor:
Could you rebase it to run tests on CI?

pfrenssen:
Rebased on latest master.

PR #180, opened 2019-09-29
Several tests are failing

pfrenssen:
When I run the tests for the `template_yarte` example I am getting the following failure:

```
---- test::test stdout ----
thread 'test::test' panicked at 'assertion failed: `(left == right)`
  left: `"text/html; charset=utf-8"`,
 right: `"text/html"`', template_yarte/src/main.rs:45:9
```

This probably went undetected since our CI pipeline doesn't run the unit tests, it only run `cargo check`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ before_script:
 
 script:
   - cargo check --all
-  - cargo test --all
+  - cargo test --all --no-fail-fast
   - |
     if [[ "$TRAVIS_RUST_VERSION" == "nightly" ]]; then
       cargo clippy
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,7 @@ before_script:
 
 script:
   - cargo check --all
+  - cargo test --all
   - |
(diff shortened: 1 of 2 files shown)


pfrenssen:
The first PR is making sure that the tests are running on Travis. It turns out that the tests for the `redis-session` example are also failing:

```
     Running target/debug/deps/redis_session-eda50a191875f194

running 1 test

test test::test_workflow ... FAILED

failures:

---- test::test_workflow stdout ----

thread 'test::test_workflow' panicked at 'called `Option::unwrap()` on a `None` value', src/libcore/option.rs:378:21
```

This is probably because Redis is not enabled in the test environment.

pfrenssen:
The `redis-session` test is now passing, but the `template_yarte` test is still failing. Let's fix this and also enable the option to run all tests regardless of any prior failures.

JohnTitor:
@svenstaro The actix seems to prefer to squash and merge commits.

svenstaro:
Note taken.

PR #85, opened 2019-03-11
Add first test in hello-world example

allevo:
As titled.

This PR introduces a test in hello-world example

PR #84, opened 2019-03-10
Add first test in json example

allevo:
I would like to start to insert some tests in the examples.
This is first PR.

PR #185, opened 2019-10-08
Add tests in examples

allevo:
Add test

This PR comes from #84 and #85

### DIFF
--- a/json/src/main.rs
+++ b/json/src/main.rs
@@ -1,5 +1,3 @@
-extern crate json;
-
 use actix_web::{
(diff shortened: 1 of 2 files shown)


json/src/main.rs
```diff
@@ -1,4 +1,3 @@
-#[macro_use]
 extern crate json;
```
JohnTitor: This line is also unnecessary.

allevo: Done

JohnTitor:
Sorry for the delay, thanks!

PR #186, opened 2019-10-08
Always send body whole and don't use transfer encoding

svenstaro:
Without doing this, body size would be unknown at the time of sending, causing awc to use transfer encoding which is probably unexpected in most cases.
This change makes behavior more obvious.

### DIFF
--- a/http-proxy/src/main.rs
+++ b/http-proxy/src/main.rs
@@ -7,7 +7,7 @@ use url::Url;
 
 fn forward(
     req: HttpRequest,
-    payload: web::Payload,
+    body: web::Bytes,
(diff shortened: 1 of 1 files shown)


svenstaro:
Please don't merge until @vmalloc has gotten this a look-over since they are the original author and I'd like to hear some thoughts on this.

JohnTitor:
Ping @vmalloc: any thoughts?

svenstaro:
Actually just today I noticed that this isn't quite correct yet either. I'll fix it. Please don't merge.

svenstaro:
Ok, I feel good about merging this now. This should also fix #188.

PR #196, opened 2019-11-27
Update Multipart example

bevanhunt:
As per conversation - reopened on master.


multipart/src/main.rs
```diff
@@ -80,18 +58,20 @@ fn index() -> HttpResponse {
...
-        App::new()
-            .data(Cell::new(0usize))
-            .wrap(middleware::Logger::default())
-            .service(
-                web::resource("/")
-                    .route(web::get().to(index))
-                    .route(web::post().to_async(upload)),
-            )
+    App::new()
+    .data(Cell::new(0usize))
```
cdbattags: anyone have an explanation for this line?

bitemyapp: https://github.com/actix/examples/blob/master/multipart/src/main.rs#L52

cdbattags: Oh, wow, that's a little unintuitive, no? `.data()` is a builder for AppState?

bitemyapp: Screenshot from 2019-12-04 12-11-59

I wouldn't say that, I think it's simply derelict.

multipart/src/main.rs
```diff
@@ -1,75 +1,53 @@
...
+use std::io::Write;
+use actix_multipart::Multipart;
+use actix_web::{middleware, web, App, Error, HttpResponse, HttpServer};
+use futures::{StreamExt};
+use std::cell::Cell;
+use serde::Deserialize;
 
 pub struct AppState {
     pub counter: Cell<usize>,
 }
```
bitemyapp: Based on https://github.com/actix/examples/pull/196#discussion_r353901959

Could you kill this struct off please? I don't think it's getting used in the example and it's misleading.

multipart/src/main.rs
```diff
@@ -1,75 +1,53 @@
-use std::cell::Cell;
-use std::fs;
-use std::io::Write;
+extern crate actix_web;
```
rofrol: >Beginning in the 2018 edition, use declarations can reference crates in the extern prelude, so it is considered unidiomatic to use extern crate.

https://doc.rust-lang.org/reference/items/extern-crates.html

multipart/Cargo.toml
```diff
@@ -1,17 +1,18 @@
...
 [dependencies]
-actix-web = "1.0.0"
-actix-multipart = "0.1.1"
-
-env_logger = "0.6"
-futures = "0.1.25"
+futures = "0.3.1"
+actix-multipart = "0.2.0-alpha.1"
+actix-web = "2.0.0-alpha.1"
+envy = "0.4.0"
```
bitemyapp: Why `envy` instead of `env_logger`? What do the other examples use?

multipart/src/main.rs
```diff
@@ -80,18 +49,19 @@ fn index() -> HttpResponse {
 
 fn main() -> std::io::Result<()> {
     std::env::set_var("RUST_LOG", "actix_server=info,actix_web=info");
-    env_logger::init();
-
+    std::fs::create_dir_all("./tmp").unwrap();
+    let config = envy::from_env::<Config>().unwrap();
+    let p =  config.port;
+    let port = concat_string!("0.0.0.0:", p);
```
bitemyapp: This isn't a port, it includes the ip address, so the naming is misleading.

I think the configurable port is a distraction here. Only one other example in this repository has a configurable port and it's https://github.com/actix/examples/blob/dc6e1cc7625179e5ab8c037d20ce6883e3f77f93/http-proxy/src/main.rs#L84 which makes sense given it's a demonstration of an http proxy.

bevanhunt: I want the example to be deployable to Heroku. Herok sets the ENV port.

bevanhunt: The IP address is preset to 0.0.0.0 so it is only setting the port.

multipart/Cargo.toml
bevanhunt: `env_logger` I will change that - thanks.

multipart/src/main.rs
bitemyapp: >I want the example to be deployable to Heroku. Heroku sets the ENV port.

@z0mbie42 okay, that's an understandable desire but what does Heroku have to do with this particular example or this repository? None of the other examples in this repository respect the port environment variable that Heroku obligates you to use.

bevanhunt: Well, I wanted to have my own repo and just put it in the readme - but they wanted the code in this repo. So here we are. Examples aren't useful unless you can deploy them. My example is a 2.0 example anyway. So probably they should be all changed. But for now, I can hard code it and do another PR later. All the examples will have to be updated to 2.0 anyway later.

bevanhunt: Thanks for the great feedback!

multipart/Cargo.toml
bevanhunt: Removed the need for env_logger/envy as hard-coded port.

cdbattags:
Good work and thanks for being patient! My last comment now is why is the semantic versioning decreasing? I would put this as a 0.3.0 and leave the name the same, no? Thoughts?

bevanhunt:
No problems - the name is the now the same, the URL is updated, and the version is 0.3.0.

fafhrd91:
is this completed?

bevanhunt:
Yes

fafhrd91:
thanks!

PR #197, opened 2019-12-07
Update Juniper example

sectore:


fafhrd91:
thanks

fafhrd91:
I merged your pr, but your code is wrong. especially around async operations. you should be careful if you use such code in prod

sectore:
@fafhrd91 Could you go into detail what might be wrong, please?

fafhrd91:
juniper is sync, so you block whole thread.

sectore:
@fafhrd91 Thanks for the fix 9ef1107 . Just realized there is another [`graphql-demo`](https://github.com/actix/examples/tree/9ef1107c0cc58ded833c052c81347ca79cc0e996/graphql-demo), which based on `Juniper` as well. How about to remove [`juniper example`](https://github.com/actix/examples/tree/9ef1107c0cc58ded833c052c81347ca79cc0e996/juniper) in favour of `graphql-demo` (this could be renamed to `juniper` or something then)? It's kinda confusing to have very similar examples for using `Actix` + `Juniper`.

PR #199, opened 2019-12-09
redo actix_redis example

bevanhunt:
Removed delete - just does write. Works with docker-compose.


fafhrd91:
I am not sure what is the reason for this change

fafhrd91:
create separate (simple) example.

PR #200, opened 2019-12-09
Update template_tera to use Tera 1.0.0

Th3Whit3Wolf:
Updated to use `template_tera` Tera 1.0.0


Th3Whit3Wolf:
It looks like the diesel example fails.

PR #202, opened 2019-12-10
disable broken projects for now

bevanhunt:
Both diesel and json are broken in master and should be disabled until they are fixed.


leizzer:
That broken example must part must be addressed not removed, IMHO

bevanhunt:
It can be dealt with later but with master broken, all other PR requests will fail.

PR #204, opened 2019-12-10
Redo Diesel example

bevanhunt:
Removed manual example (simplified) & got working on docker-compose with Postgres.


fafhrd91:
this is example, it exists to showcase how to do things with comments. why did you remove actual example?

fafhrd91:
same as other prs

leizzer:
Examples should serve as documentation, reducing the need for people to reach out on other channels.
I also don't understand why taking out Sqlite and depending on docker is better

PR #205, opened 2019-12-10
redo json example - make simple

bevanhunt:
The current example is broken and too complex for a simple example.


leizzer:
If it is broken that is one thing and we need to fix it.
And the complexity you mention is in fact various examples in one file for different needs, each one has its own purpose. I feel like you are taking out examples.

fafhrd91:
this example shows different type json handling. what is the reason for removing?

bevanhunt:
The examples are too complex for a simple hello world for JSON.  Sure we can have another more complicated example(s) in another folder. However, I want to know quickly how to do a JSON hello world. It is not trivial to parse through the code with all the current examples to figure this out quickly.

fafhrd91:
if it complex then create new one, json_helloworld or json_basic.

PR #207, opened 2019-12-10
Fix examples of diesel and json

JohnTitor:


leizzer:
this was failing to me earlier today! great work

diesel/src/main.rs
```diff
@@ -60,7 +60,7 @@ async fn add(
 
 /// This handler manually parse json object. Bytes object supports FromRequest trait (extractor)
 /// and could be loaded from request payload automatically
-async fn index_add(body: Bytes, pool: web::Data<Pool>) -> Result<HttpResponse, Error> {
```
fafhrd91: this won't work. `let mut body = BytesMut::new();` must be removed

JohnTitor: Fixed

PR #221, opened 2019-12-26
Fix logging in basics example

kamnxt:
Basics example had a semicolon instead of a comma, causing logging not to work properly.


fafhrd91:
thanks!

PR #224, opened 2020-01-06
Fix path name on /examples/basics

BC-M:
A wrong route is given in the readme file. The correct route is /async-body/{name}.


JohnTitor:
Thanks!

PR #228, opened 2020-01-11
Use async-std in multipart example and treat all errors

torfmaster:
# Summary

In this PR I use `async_std` instead of blocking. Moreover, I treat all errors. I hope this makes the code more streamlined.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ members = [
    "juniper",
    "middleware",
    "multipart",
+   "multipart-async-std",
    "openssl",
 #   "protobuf",
    "r2d2",
--- a/multipart-async-std/.gitignore
+++ b/multipart-async-std/.gitignore
@@ -1,0 +1,3 @@
+/target
+**/*.rs.bk
+/tmp
--- a/multipart-async-std/Cargo.toml
+++ b/multipart-async-std/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "multipart-async-std-example"
+version = "0.3.0"
+authors = ["Bevan Hunt <bevan@bevanhunt.com>"]
+edition = "2018"
+license = "MIT"
+description = "Simple file uploader in Actix Web with Async/Await"
+keywords = ["actix", "actix-web", "multipart"]
+repository = "https://github.com/actix/examples"
+readme = "README.md"
+
+[dependencies]
+futures = "0.3.1"
+actix-multipart = "0.2.0"
+actix-web = "2.0.0"
+actix-rt = "1.0.0"
+async-std = "1.4.0"
--- a/multipart-async-std/LICENSE
+++ b/multipart-async-std/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) [2019] [Bevan Hunt]
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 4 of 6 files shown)


JohnTitor:
Test has been fixed on master, could you drop the commit?

torfmaster:
Fixed.

multipart/src/main.rs
```diff
@@ -1,25 +1,26 @@
-use std::io::Write;
-
 use actix_multipart::Multipart;
 use actix_web::{middleware, web, App, Error, HttpResponse, HttpServer};
+use async_std;
```
JohnTitor: This line is unnecessary.

fafhrd91:
let's create new example for async-std

JohnTitor:
Ah, okay, @torfmaster could you create a new example as @fafhrd91 mentioned?

multipart/src/main.rs
torfmaster: fixed.

torfmaster:
Okay, I created a new example.

JohnTitor:
Thanks!

cdbattags:
I'm curious how the event loop is handled with this under the hood.

Using async_std in tandem with tokio means multiple runtimes, no?

cdbattags:
@actix/contributors, can any of y'all shed some light on how this works under the hood?

> I'm curious how the event loop is handled with this under the hood.
> 
> Using async_std in tandem with tokio means multiple runtimes, no?

Dowwie:
@cdbattags please reserve the callout to all contributors for matters of importance

cdbattags:
@Dowwie, uhh, what does "matters of importance" mean here?

If people use `async_std` like this they'll need to know that it's going to use a completely separate runtime than "app"/tokio which is quite disconcerting.

cdbattags:
This is why @JohnTitor put out a blast to us 3 days ago about potentially trimming the team.

jwdeitch:
@cdbattags - sorry, nieve question, is the tokio runtime running in the context of this example?

cdbattags:
@jwdeitch `actix-web` uses tokio for runtime under the hood and `async_std` uses a different runtime. By calling an `async_std` method it will create yet another event loop on the same master thread, I believe, but I want to discuss in the context of this change.

tyranron:
@cdbattags @jwdeitch 

> I'm curious how the event loop is handled with this under the hood.

> sorry, nieve question, is the tokio runtime running in the context of this example?

Well, `tokio` runtime is definitely running here, because `actix_rt` uses it under the hood, so thw whole `actix-web` server runs in `actix_rt::main` as usual.

Now, how `async_std` work here? Let's look a bit into sources...  
For example, [`async_std::fs::create_dir`](https://docs.rs/async-std/1.4.0/async_std/fs/fn.create_dir.html) runs [`async_std::spawn_blocking`](https://docs.rs/async-std/1.4.0/async_std/task/fn.spawn_blocking.html) [under-the-hood](https://docs.rs/async-std/1.4.0/src/async_std/fs/create_dir.rs.html#36-43). So, technically, it starts the whole `async_std` runtime in separate threads, spawns the future there moving it into `async_std` executor and once it finished returns the result to the current thread.

tyranron:
From the documentation of [`async_std::task::spawn_blocking`](https://docs.rs/async-std/1.4.0/async_std/task/fn.spawn_blocking.html) it seems to work near the same way as [`actix_threadpool::run`](https://docs.rs/actix-threadpool/0.3.1/actix_threadpool/fn.run.html) does.

cdbattags:
My concern is that if we're showing this as a certified example then the architecture of the whole setup changes.

Where would we like to document this kind of stuff?

cdbattags:
Ideally, all futures spawned by usages of `async_std` won't hang around too long but a _single_ usage of `async_std` will forever leave around this extra runtime without any sort of cleanup.

Dowwie:
I concur with @cdbattags that this example is problematic.  actix/examples is an official repo showing acceptable patterns, not anti-patterns.  With that given, @cdbattags are you going to submit a PR to fix?

tyranron:
@cdbattags 

> will forever leave around this extra runtime without any sort of cleanup.

Yup, the same happens with `actix-threadpool`.

cdbattags:
I'd say this isn't necessarily an "anti-pattern" but there's no great solution yet to mixing `async_std` and `tokio` as of yet.

Someday soon if the traits start lining up then it won't be a concern but for now we need to advertise the weirdness of this under the hood.

PR #230, opened 2020-01-12
simple-auth-server - use derive feature from serde 

rofrol:
https://github.com/serde-rs/serde/issues/1441#issuecomment-445481084


JohnTitor:
Thanks!

PR #233, opened 2020-01-12
Update juniper example

rofrol:
>The `graphql_object!` macro is deprecated and will be removed soon.

https://github.com/graphql-rust/juniper/blob/9e849fe5122021af285e9cfd6e526123e0ad15f6/juniper/src/macros/object.rs#L5


JohnTitor:
Thanks!

PR #240, opened 2020-01-23
add self shutdown route example

robjtede:
in response to actix/actix-web#1315

### DIFF
--- a/self-shutdown-route/src/main.rs
+++ b/self-shutdown-route/src/main.rs
@@ -1,0 +1,56 @@
+use std::{sync::mpsc, thread};
+
+use futures::executor;
+
+use actix_web::{get, middleware, post, web, App, HttpResponse, HttpServer};
+
+#[get("/hello")]
+async fn hello() -> &'static str {
+    "Hello world!"
+}
+
+#[post("/stop")]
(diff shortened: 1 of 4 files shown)


JohnTitor:
It'd be nice if the example had README with a brief overview and a link to the issue.

Dowwie:
I realize this was in line with the original request but the thought of having the power to shut down a server via an http request sends chills down my spine.  What do you think about recommending against this in the README and instead using local signal handling?

JohnTitor:
:+1:

self-shutdown-route/src/main.rs
```diff
@@ -0,0 +1,56 @@
...
+            .data(stopper)
+            .wrap(middleware::Logger::default())
+            .service(hello)
+            .service(stop)
+    })
+    .bind(&bind)?
+    .run();
+
+    // clone the Server handle
+    let srv = server.clone();
```
azadsagar, line 45: Why do I get : no method named `clone` found for struct `actix_web::dev::Server` in the current scope ?

robjtede: because this example has since been reworked for actix web v4, see the master branch's shutdown-server example

PR #243, opened 2020-01-29
add an actix-web 2.0 project from community

danloh:

PR #244, opened 2020-01-30
Add dataloader example

jayy-lmao:
An example for using dataloaders via context with actix-web and juniper async-await branch.

https://github.com/jayy-lmao/rust-graphql-docker

If you'd prefer, also happy to just have it in the community section.

### DIFF
--- a/juniper_dataloaders/.gitignore
+++ b/juniper_dataloaders/.gitignore
@@ -1,0 +1,2 @@
+/target
+**/*.rs.bk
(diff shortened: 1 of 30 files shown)


JohnTitor:
I think this shouldn't be added as code since it's too large to be added as an example. Could you add a link instead like #243?

jayy-lmao:
> I think this shouldn't be added as code since it's too large to be added as an example. Could you add a link instead like #243?

Will do :)

PR #248, opened 2020-02-01
Fix templates path

### LINKED ISSUE #247: Todo example does not resolve template locations correctly
https://github.com/actix/examples/blob/741a5309632d88f3ebd9798d73db22247b9f35df/todo/src/main.rs#L36

From:>
```rust
let templates: Tera = Tera::new("/templates/**/*").unwrap();
```

To:> 
```rust
let templates = Tera::new(concat!(env!("CARGO_MANIFEST_DIR"), "/templates/**/*")).unwrap();
```

elliotekj:
Hi,

This PR fixes the leading slash error pointed out in #247.

Closes #247.

### DIFF
--- a/todo/src/main.rs
+++ b/todo/src/main.rs
@@ -33,7 +33,7 @@ async fn main() -> io::Result<()> {
     let app = move || {
         debug!("Constructing the App");
 
-        let templates: Tera = Tera::new("/templates/**/*").unwrap();
+        let templates: Tera = Tera::new("templates/**/*").unwrap();
 
         let session_store = CookieSession::signed(SESSION_SIGNING_KEY).secure(false);
 


todo/src/main.rs
```diff
@@ -33,7 +33,7 @@ async fn main() -> io::Result<()> {
     let app = move || {
         debug!("Constructing the App");
 
-        let templates: Tera = Tera::new("/templates/**/*").unwrap();
+        let templates: Tera = Tera::new("templates/**/*").unwrap();
```
robjtede: Use the cargo manifest relative path since it allows the example to be run from the root with `cargo run --bin actix-todo` (albeit without static files loading correctly).

```suggestion
        let templates: Tera = Tera::new(concat!(env!("CARGO_MANIFEST_DIR"), "/templates/**/*")).unwrap();
```

elliotekj: Happy to make the change but `cargo run --bin actix-todo` panics either way. For it to run, it has to load `DATABASE_URL` from a .env file that isn't present at the top level of the project.

robjtede: DATABASE_URL can be imput as a regular env var as well, eg `DATABASE_URL=x@localhost/db cargo run --bin actix-todo`. Maybe for other reasons it's best to leave it as is so folks running the example arent confused when part of it works and part doesnt.

JohnTitor:
Thanks!

PR #250, opened 2020-02-05
multipart + s3 example

cheolgyu:
multipart -> s3 upload
multipart-data -> struct


multipart-s3/src/main.rs
```diff
@@ -0,0 +1,100 @@
...
+
+    HttpResponse::Ok().body(html)
+}
+
+#[actix_rt::main]
+async fn main() -> std::io::Result<()> {
+    std::env::set_var("RUST_LOG", "actix_server=info,actix_web=info");
+    std::env::set_var("AWS_ACCESS_KEY_ID", "your_key");
+    std::env::set_var("AWS_SECRET_ACCESS_KEY", "your_key");
+    std::env::set_var("AWS_S3_BUCKET_NAME", "your_key");
```
elliotekj: Requiring source code edits to get an example running isn't ideal. I'd suggest loading these via a .env file.

multipart-s3/src/utils/upload.rs
```diff
@@ -0,0 +1,120 @@
...
+        let mut field: Field = item.expect(" split_payload err");
+        let content_type = field.content_disposition().unwrap();
+        let name = content_type.get_name().unwrap();
+        if name == "data" {
+            while let Some(chunk) = field.next().await {
+                data = chunk.expect(" split_payload err chunk");
+            }
+        } else {
+            if content_type.get_filename() != None {
+                let filename = content_type.get_filename().unwrap();
```
elliotekj: This can be shortened to:

```rust
if let Some(filename) = content_type.get_filename() {
```

multipart-s3/src/main.rs
cheolgyu: ok

multipart-s3/src/utils/upload.rs
cheolgyu: @elliotekj ok!

robjtede:
Hey, thanks for the contribution! For ease of reading can we have the HTML index doc in a separate file and included with `include_str!`. Also the readme could use some work, see here for a great example to follow: https://github.com/actix/examples/tree/master/diesel

leizzer:
Hi @robjtede! Looks like @cheolgyu did the changes that you requested. 

Do you consider we can merge it?

robjtede:
I didn't merge incase they split out the HTML file but I'm concerned enought to let this go on longer.

leizzer:
I understand your concern, but it's not the only example with the html in the same file.
And it's a good example that I would like to have :)

multipart-s3/src/utils/upload.rs
```diff
@@ -0,0 +1,124 @@
...
+    tmp_files: Vec<Tmpfile>,
+    s3_upload_key: String,
+) -> Result<Vec<UplodFile>, Error> {
+    let mut arr: Vec<UplodFile> = Vec::new();
+    let mut iter = tmp_files.iter();
+    let mut index = 0;
+    // iterate over multipart stream
+    while let Some(item) = iter.next() {
+        let mut tmp_file: Tmpfile = item.clone();
+        tmp_file.s3_upload_and_tmp_remove(s3_upload_key.clone());
```
vmalloc: It seems to me this function is not truly async, as it will block during the upload phase. Is this by design? I think such usage is discouraged in actual production apps as it will block the current worker for potentially long durations.

vmalloc:
I think this example should be fixed to demonstrate correct delegation to asynchronous uploading. As it is, it advertises a usage that can cause issues in real-world apps.

@robjtede what do you think?

multipart-s3/src/utils/upload.rs
leizzer: `put_object` is async if I remember it correclty.
Sorry that I didn't see your comment on time. I'll give it another look when I get home.

robjtede:
As much as possible, the examples should display best practices at the same time as focussing on their primary feature. So, I'd say that yes the `.sync()` call that blocks is problematic.

multipart-s3/src/utils/upload.rs
vmalloc: It is async, but under the hood you call it with sync(), making it synchronous. In general, an async fn without any awaits in it should be considered a red flag.

vmalloc:
@cheolgyu can you please submit a PR to address this issue? Otherwise I think this example is best be removed for now or replaced with something not involving s3 to focus on the things that work correctly

cheolgyu:
> @cheolgyu can you please submit a PR to address this issue? Otherwise I think this example is best be removed for now or replaced with something not involving s3 to focus on the things that work correctly

@vmalloc Okay.
I'll fix it.

cheolgyu:
고마워 @vmalloc
 https://github.com/actix/examples/pull/252

vmalloc:
@cheolgyu Thanks for the quick fix!

RaldyDaldy:
Reply

Op vr 3 apr. 2020 21:24 schreef robbieS <notifications@github.com>:

PR #251, opened 2020-02-06
Update yarte version to 0.6.0

botika:
I have updated to the new version and introduced the new features in the templates.

In this new version, [partial-block](https://handlebarsjs.com/guide/partials.html#partial-blocks) support, a first version of html minifier and partial and partial-block recursion support (such as Rust macros but evaluating the expressions, conditionals, loops, ..) have been introduced.


JohnTitor:
Thanks!

PR #252, opened 2020-02-11
update rusoto_core = "0.43.0-beta.1"

cheolgyu:


robjtede:
LGTM. Figures that the original problem was that latest non-beta rusoto wasn't std::future.

JohnTitor:
Thanks!

PR #260, opened 2020-02-20
Add `docker_sample` to examples collection

curiousdev:
Build image:

```shell
  docker build -t docker_sample --force-rm --no-cache -f  Dockerfile .
```

Run image:

```shell
  
  echo "== start sample_docker"
  docker run -d -p 5000:5000 docker_sample &
  docker ps

  echo "== wait 3s for startup"
  sleep 3s

  echo "== curl both routes"
  curl http://localhost:5000
  curl http://localhost:5000/again 

```

### DIFF
--- a/docker_sample/Dockerfile
+++ b/docker_sample/Dockerfile
@@ -1,0 +1,15 @@
+FROM rust:1 as builder
+
+COPY . .
+
+RUN cargo build --release
+
+FROM rust:1-slim-stretch
+
+COPY --from=builder /target/release/docker_sample .
+
+RUN ls -la /docker_sample
+
+EXPOSE 5000
+
+ENTRYPOINT ["/docker_sample"]
--- a/docker_sample/src/main.rs
+++ b/docker_sample/src/main.rs
@@ -1,0 +1,30 @@
+#[macro_use]
+extern crate actix_web;
+
+use actix_web::{web, App, HttpResponse, HttpServer, Responder};
+
+#[get("/")]
+async fn index() -> impl Responder {
+    println!("GET: /");
+    HttpResponse::Ok().body("Hello world!")
+}
+
+#[get("/again")]
+async fn again() -> impl Responder {
+    println!("GET: /again");
+    HttpResponse::Ok().body("Hello world again!")
+}
+
+#[actix_rt::main]
+async fn main() -> std::io::Result<()> {
+    println!("Starting actix-web server");
+
+    HttpServer::new(|| {
+        App::new()
+            .service(index)
+            .service(again)
+    })
+    .bind("0.0.0.0:5000")?
+    .run()
+    .await
+}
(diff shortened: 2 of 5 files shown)


curiousdev:
Please let me know if I've missed anything. I'd like to get this added to the examples collection.

docker_sample/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+COPY . .
+
+# Sets the environment variable for the cargo build command that follows.
+ENV PKG_CONFIG_ALLOW_CROSS=1
+
+RUN cargo build --release
+
+FROM rust:1.41.0-stretch
+
+#RUN apk --no-cache add ca-certificates
```
Jonathas-Conceicao: If this command is not required I think it should be left here on the Dockerfile.

docker_sample/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM rust:1.41.0-stretch as builder
```
Jonathas-Conceicao: Not sure about this one but: wouldn't it be better to use the `1` tag of rust image? Actix itself will be maintained to work with stable releases of Rust, so it would be nice if the Dockerfile don't need to be updated in case we raise the minimal required Rust version to a new stable release.

docker_sample/src/main.rs
```diff
@@ -0,0 +1,25 @@
...
+}
+
+#[actix_rt::main]
+async fn main() -> std::io::Result<()> {
+    println!("Starting actix-web server");
+
+    HttpServer::new(|| {
+        App::new()
+            .route("/", web::get().to(index))
+            .route("/again", web::get().to(index2))
```
Jonathas-Conceicao: How about using the codegen macros here  with `.service` instead of `.route` directly? The [basics](https://github.com/actix/examples/blob/master/basics/src/main.rs#L17) example does it like that and I personally think this improves code readability, I think this is the default way of implementing routing since 1.0.

Jonathas-Conceicao:
Thanks!

Dowwie:
@Jonathas-Conceicao What about this PR warranted it's submission to the examples?

Jonathas-Conceicao:
@Dowwie It is my understanding that some developers start new projects straightaway using Docker. Even though we have some projects under the "from community" that setup the environment using Docker, I think it's a good idea to have a minimalist example that can serve as a starting point to someone using actix-web.

Dowwie:
@Jonathas-Conceicao I agree with you that it would be good idea to have a build/distribution example.  It would be *great* to have a comprehensive one.  The problem is that this example is far too minimal as to be useful for real world use.  Rust Take a look at this project as a counter-example: https://github.com/olxgroup-oss/dali

Here's a list of relevant articles: https://readrust.net/devops-and-deployment

PR #264, opened 2020-03-02
Add a basic casbin example for actix-web.

PsiACE:
Casbin-RS is a powerful and efficient open-source access control library for Rust projects. It provides support for enforcing authorization based on various access control models.

Here is a basic example, hope to help those who want to use it.

### DIFF
--- a/casbin/src/main.rs
+++ b/casbin/src/main.rs
@@ -14,7 +14,7 @@ async fn success(
     println!("{:?}", req);
     assert_eq!(vec!["data2_admin"], e.get_roles_for_user("alice", None));
 
-    HttpResponse::Ok().body(format!("Num of requests: {}", 1))
+    HttpResponse::Ok().body("Success: alice is data2_admin.")
 }
 
 async fn fail(enforcer: web::Data<RwLock<Enforcer>>, req: HttpRequest) -> HttpResponse {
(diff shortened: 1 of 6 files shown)


casbin/src/main.rs
```diff
@@ -0,0 +1,58 @@
...
+/// simple handle
+async fn success(
+    enforcer: web::Data<RwLock<Enforcer>>,
+    req: HttpRequest,
+) -> HttpResponse {
+    let mut e = enforcer.write().unwrap();
+    println!("{:?}", req);
+    assert_eq!(vec!["data2_admin"], e.get_roles_for_user("alice", None));
+
+    HttpResponse::Ok().body(format!("Num of requests: {}", 1))
```
JohnTitor: What's the intention of this `format!` here?

PsiACE: Obviously, just output something to indicate that this access request is valid.

JohnTitor: So, it's unnecessary to use `format!` and hard-coded value, it should be just a str.

PsiACE: Got it.

I think another thing worth discussing is whether I should use `assert_eq!`. If it fails, it may cause the program to not work properly.

JohnTitor: > I think another thing worth discussing is whether I should use assert_eq!. If it fails, it may cause the program to not work properly.

I'm fine with that since it's just an example and the purpose is obvious. But feel free to extract them.

PsiACE:
Please check and merge (if possible).

PR #268, opened 2020-03-08
New example: Return JSON decode errors to a client

Stigjb:
A demonstration of how you can return something more informative to the client than an empty HTTP 400 in cases where deserializing a JSON payload fails.

Please let me know if the coding style, documentation, naming or anything else can be improved, I am not very experienced with programming in Rust yet!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ members = [
    "hello-world",
    "http-proxy",
    "json",
+   "json_decode_error",
    "json_error",
    "jsonrpc",
    "juniper",
--- a/json_decode_error/Cargo.toml
+++ b/json_decode_error/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "json_decode_error"
+version = "0.1.0"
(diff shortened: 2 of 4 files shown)


Firstyear:
Instead of this being an example here, could this actually be a docstring example instead? That would make more sense to help keep the information localised and discoverable.

Stigjb:
Do you mean that it would fit better in a different repository (where?), or that the examples from the README should be docstrings examples in the source file?

robjtede:
I'm in favor of merging this. It's good to have a complete example of this working together and some people may miss the link with JsonConfig since the docs for Json only give it a brief mention.

Though, adding to the doc-comment examples is a good idea, too.

Firstyear:
So that when someone goes to this link, they see your example: 

https://docs.rs/actix-web/2.0.0/actix_web/web/struct.JsonConfig.html

Stigjb:
I submitted a PR to https://github.com/actix/actix-web/pull/1418 now, with some differences from this, such as removing the 422 response

PR #270, opened 2020-03-11
add AWS_REGION env variable and add _ prefix

ggomagundan:
Propose update some minor code 

- Add `AWS_REGION` variable on `.env` file
  - Get from [AWS Regions](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html)
- Add `_` prefix unused variables
- Remove unused `import`


multipart-s3/src/utils/s3.rs
```diff
@@ -1,8 +1,7 @@
 use crate::rusoto_s3::S3;
-use rusoto_core::{Region, RusotoError, RusotoFuture};
+use rusoto_core::{Region};
```
JohnTitor:
```suggestion
use rusoto_core::Region;
```

multipart-s3/src/utils/s3.rs
```diff
@@ -13,20 +12,42 @@ pub struct Client {
...
-        Client {
-            region: region.to_owned(),
-            s3: S3Client::new(region),
-            bucket_name: std::env::var("AWS_S3_BUCKET_NAME").unwrap(),
-        }
+      Client {
+          region: region.to_owned(),
+          s3: S3Client::new(region),
+          bucket_name: std::env::var("AWS_S3_BUCKET_NAME").unwrap(),
+      }
```
JohnTitor: Why did you change indent size?

multipart-s3/src/utils/s3.rs
```diff
@@ -13,20 +12,42 @@ pub struct Client {
...
+        "eu-west-3" => Region::EuWest3,
+        "me-south-1" => Region::MeSouth1,
+        "sa-east-1" => Region::SaEast1,
+        "us-east-1" => Region::UsEast1,
+        "us-east-2" => Region::UsEast2,
+        "us-west-1" => Region::UsWest1,
+        "us-west-2" => Region::UsWest2,
+        // Default
+        _ => Region::ApNortheast2
+      };
```
JohnTitor: We should use `Region::default()` here. ref: https://docs.rs/rusoto_core/0.43.0/rusoto_core/enum.Region.html#default

ggomagundan: so Thanks.I will fix it

multipart-s3/src/utils/s3.rs
ggomagundan: okay. My editors depends other langs(maybe `js`) configurations. I will change 2 to 4 indent size

multipart-s3/src/utils/s3.rs
JohnTitor: `Region::default()` will attempt to read `AWS_REGION` env so we don't need this match expression.

JohnTitor:
Hey could you make a progress on this? I'd like to clean-up that example but this PR blocks me.

multipart-s3/src/utils/s3.rs
ggomagundan: So sorry to late fix.

JohnTitor: No worries, thanks for the work!

PR #278, opened 2020-03-23
Add async-graphql example

sunli829:


robjtede:
Hey @sunli829, can you update this to actix-web v3 so we can look at merging it.

sunli829:
> Hey @sunli829, can you update this to actix-web v3 and rebase on master so we can look at merging it.

Okay, I upgraded to actix-web 3.0, thank you for your reply.

sunli829:
I have updated the code, please review. 😁 @robjtede

Cargo.toml
```diff
@@ -52,4 +52,6 @@ members = [
    "websocket-chat",
    "websocket-chat-broker",
    "websocket-tcp-chat",
+   "docker_sample",
+   "async-graphql-demo"
```
JohnTitor, line 56: `docker_sample` here should be removed as duplicated and `async-graphql-demo` should be ordered alphabetically.

async-graphql-demo/src/main.rs
```diff
@@ -0,0 +1,38 @@
...
+
+async fn index_playground() -> Result<HttpResponse> {
+    Ok(HttpResponse::Ok()
+        .content_type("text/html; charset=utf-8")
+        .body(playground_source(
+            GraphQLPlaygroundConfig::new("/").subscription_endpoint("/"),
+        )))
+}
+
+#[actix_rt::main]
```
robjtede, line 21: This is the recommended way now and can remove the actix-rt dependency.

```suggestion
#[actix_web::main]
```

sunli829: Thank you! 🙂

robjtede:
This needs a readme similar to other ones ([here's a good reference](https://github.com/actix/examples/blob/master/diesel/README.md)) just to quickly explain the example and show folks what endpoints they can call.

robjtede (approved):
LGTM, thanks :)

PR #300, opened 2020-04-18
Update dependencies for casbin example.

PsiACE:


casbin/src/main.rs
```diff
@@ -27,8 +26,7 @@ async fn fail(enforcer: web::Data<RwLock<Enforcer>>, req: HttpRequest) -> HttpRe
 
 #[actix_rt::main]
 async fn main() -> io::Result<()> {
-    std::env::set_var("RUST_LOG", "info");
-    std::env::set_var("LOGE_FORMAT", "target");
```
JohnTitor, line 31: Hm, I think it's simpler to use `set_var()`?

PsiACE: You are right, done!

PR #301, opened 2020-04-20
fix thread block in diesel pack and compatibility in async_pg pack

Icatream:
I don't know how to split these two commits.

The first one is about `pool.get()` block the current thread, i moved it into `web::block` block.

The second commit is a compatibility for windows.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ members = [
    "simple-auth-server",
    "state",
    "static_index",
+   "stop_handling",
    "template_askama",
    "template_handlebars",
    "template_tera",
--- a/async_pg/Cargo.toml
+++ b/async_pg/Cargo.toml
@@ -8,7 +8,6 @@ workspace = ".."
 [dependencies]
 actix-rt = "1.0.0"
 actix-web = "2.0.0"
-actix-server = "1.0.2"
 config = "0.10.1"
 deadpool-postgres = "0.5.5"
 derive_more = "0.99.3"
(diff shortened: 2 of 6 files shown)


Icatream:
The single quote was formatted by vscode plugin [toml-formatter](https://marketplace.visualstudio.com/items?itemName=Iceyer.toml-formatter).
Is there any other plugin should i to use?

robjtede:
> Is there any other plugin...

Don't use one, personally. Someone suggested that plugin also removes comments form toml files so maybe not the best choice. None of the other toml files in this repo have single quotes.

Icatream:
That plugin do removes comments, and i haven't find another plugin can format toml, i'll write toml manual now.

Icatream:
So, just create a new pack, write the stop handling in it?

Icatream:
I have wrote a `stop-handling` package and test it.
The stop handling works as it expected in linux.
But in windows, it won't print `SIGINT Received.`, just stopped.
I've followed the explanation in `tokio::signal::windows`.
The only thing worked, is it compiled in windows, as the `async_pg` can't be compiled before adding the compatibility.
Do i need to pull it?

Icatream:
I've got confused by the "ctrl break" and "ctrl c".
This time, it works fine.

Icatream:
I've forgot why do i did this stop handling, seems that it used to have a unix stop handling in `async_pg`, and prevent compile in windows. May be i had some confusing with memory.

Dowwie:
async_pg *did* in fact contain a custom implementation for sigint shutdown and it was a mistake

robjtede:
closing as stale

PR #302, opened 2020-04-20
Add: MongoDB example

mehmetsefabalik:


cdbattags:
Before we add this I'd love a sample DB import/export included with the example.

Also, can we cleanup this README a bit? [rust-mongodb-example/README.md at master · mehmetsefabalik/rust-mongodb-example](https://github.com/mehmetsefabalik/rust-mongodb-example/blob/master/README.md)

cdbattags:
I'm gonna guess that the autocannon arg is wrong here: https://github.com/mehmetsefabalik/rust-mongodb-example/blame/master/README.md#L56?

SachinMaharana:
It seems r2d2-mongodb being used here is deprecated. mongodb crate internally does connection pooling. Can we update to reflect this? Follwing thread explains this in more detail.
https://gitlab.com/petoknm/r2d2-mongodb

https://github.com/mongodb/mongo-rust-driver/pull/109#issuecomment-569751305

Issue #295, opened 2020-04-14
There are two graphQl example. both using Juniper. Should one be removed?


kunjee17:
Hi, First of all. Thanks for all examples. As some cleaning up of repo is going on (figure out from commit messages). It would be great if graphQL example can be merged. As of now there are two examples in repo and both are using Juniper. I haven't look into details but if there is no significant difference then once should be and can be removed.

JohnTitor:
So, the `graphql-demo` example has more detail than the `juniper` example, we could remove the later.
But if someone wants to have simple example, we could just rename them to clarify (I guess `graphql-demo` should be `juniper-r2d2` or something else).

kunjee17:
@JohnTitor  that would be great. Because I m new to actix and rust. So, at least I know where to start and where I can find advanced example.

PR #304, opened 2020-04-22
Another multipart example

kantimam:
pretty basic multipart form example. Mostly based of the multipart+s3 example just more basic and hopefully easier to understand for potatoes like me.


multipart-basic/src/main.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    println!("bytes={:#?}", form);
+    
+    println!("files={:#?}", files);
+    
+    Ok(HttpResponse::Ok().into())
+}
+
+fn index() -> HttpResponse {
+    let html = r#"<html>
```
robjtede, line 20: include the HTML from an external file using something like include_str! macro

multipart-basic/src/main.rs
```diff
@@ -0,0 +1,79 @@
...
+    HttpResponse::Ok().body(html)
+}
+
+#[actix_rt::main]
+async fn main() -> std::io::Result<()> {
+    
+    std::env::set_var("RUST_LOG", "actix_server=info,actix_web=info");
+    std::fs::create_dir_all("./files").unwrap();
+
+    let ip = "0.0.0.0:3000";
```
robjtede, line 67: most examples start on `127.0.0.1:8080`

multipart-basic/src/main.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    HttpServer::new(|| {
+        App::new().wrap(middleware::Logger::default()).service(
+            web::resource("/")
+                .route(web::get().to(index))
```
robjtede, line 72: replace with simpler #[get] and #[post] macros on the handlers

multipart-basic/src/utils/payload_handler.rs
```diff
@@ -0,0 +1,99 @@
...
+    while let Some(item) = payload.next().await {
+        let mut field: Field = item.expect(" split_payload err");
+        let content_type = field.content_disposition().unwrap();
+        let name = content_type.get_name().unwrap();
+        if name != "file" {
+            while let Some(chunk) = field.next().await {
+                let data = chunk.expect("split_payload err chunk");
+                /* convert bytes to string and print it  (just for testing) */
+                if let Ok(s)=str::from_utf8(&data){
+                    println!("{:?}", s);
```
robjtede, line 51: if we're going to include debug prints, can we use log + env_logger and use debug! macro here

multipart-basic/src/utils/payload_handler.rs
```diff
@@ -0,0 +1,99 @@
...
+    title: String,
+    description: String, 
+    count: u32
+}
+
+pub async fn split_payload(payload: &mut Multipart) -> (FormOut, Vec<UploadedFiles>) {
+    let mut files: Vec<UploadedFiles> = Vec::new();
+
+    /* fill with default values for now */
+    let mut form : FormOut=FormOut {
```
robjtede, line 36: good practice to have an impl Default when you need to init like this

JohnTitor:
My concern is that multiple examples may make us confused, like #295. And we have already three multipart examples. I'm not sure we really need this one, @actix/contributors thoughts?

multipart-basic/src/utils/payload_handler.rs
```diff
@@ -0,0 +1,99 @@
...
+}
+
+#[derive(Deserialize, Serialize, Debug)]
+pub struct FormOut {
+    title: String,
+    description: String, 
+    count: u32
+}
+
+pub async fn split_payload(payload: &mut Multipart) -> (FormOut, Vec<UploadedFiles>) {
```
robjtede, line 32: this function could use more comments imo

robjtede:
On wether to include this example at all: actually hadnt spotted the existing basic multipart example. That one seems simpler than this even.

kantimam:
ya fair enough. Kinda struggled to figure out how to get the text fields since both multipart examples focus on files mostly. Maybe add another input apart from only files to one of them?

robjtede:
> Maybe add another input apart from only files to one of them?

Sounds good, if you want to add some text handling to that example a PR would be welcome IMO.

kantimam:
oki :) will do that instead

kantimam:
btw the actix-form-data example here https://crates.io/crates/actix-form-data is no longer valid? since i try to get it to work for like 2 days now.

JohnTitor:
> btw the actix-form-data example here https://crates.io/crates/actix-form-data is no longer valid? since i try to get it to work for like 2 days now.

It isn't a part of the actix project and depends on actix-web v1.

PR #309, opened 2020-04-29
feature-a restful service example

zhangyebai:
a micro service base on rust, actix, actix_*, deadpool, tokio for restful interface.
example for: 
- query parameters
- async pg
- log
- muti workers

### DIFF
--- a/restful_service/README.MD
+++ b/restful_service/README.MD
@@ -1,0 +1,129 @@
+
+### 世间唯你好, 余者皆枯骨
+```
+    Rust写的一个Restful接口微服务, 尽量符合springboot的风格, 使用middleware处理aop, 全局异常处理器等.
+    - actix
+    - actix_web
+    - deadpool
+    - tokio-pg
+```
+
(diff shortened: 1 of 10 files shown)


restful_service/README.MD
```diff
@@ -0,0 +1,129 @@
+
+### 世间唯你好, 余者皆枯骨
+```
+    Rust写的一个Restful接口微服务, 尽量符合springboot的风格, 使用middleware处理aop, 全局异常处理器等.
+    - actix
+    - actix_web
+    - deadpool
+    - tokio-pg
+```
```
JohnTitor, line 9: Please use English instead.

robjtede:
> So, at a glance, we have similar example (`async_pg`) and I'd improve it rather than add this one.

Agreed, I like the idea of including `deadpool` in an example though.

JohnTitor:
Closing as inactivity, feel free to re-open with updates. Thanks!

PR #317, opened 2020-05-17
Improve the static files example.

### LINKED ISSUE #316: static_index example needs some love
The code from the `static_index` example has been taken from the `websocket` example but is incomplete and results in a website that doesn't work. Also the README is empty, and it would maybe be a good idea to demonstrate a few of the features that are offered, like enabling a directory listing on a certain folder.

pfrenssen:
* Replace the broken websocket example with a simple static HTML page.
* Show how to serve a second directory which is located inside the main directory but has file listing enabled.
* Demonstrate that a hierarchy of static files can be served by placing the JS and CSS files in their respective subfolders.
* Document how to use the example in the README.

Fixes #316


pfrenssen:
The failure is  because a new version of `tokio-util` is being pulled in, this is also affecting the master branch. I will make an issue.

PR #323, opened 2020-05-19
Demonstrate how to access templating engines in app_data() to render HTML error pages

pfrenssen:
The `handlebars` and `tera` templating engines are using app data to store their structs and these are passed on to handlers as an extractor. It is however not clear how to access these in error handlers.

This PR adds a practical example on how a 404 page can be returned as a HTML response using the templating engine extracted from `app_data()` inside an error handler.


JohnTitor:
Looks good! It's unrelated to this PR but it'd be great if you could also move README of handlebars example from `template_handlebars/src/README.md` to `template_handlebars/README.md`.

PR #331, opened 2020-06-13
Add bare-bones WS server for autobahn testing

mlodato517:
**This PR**
Hopefully creates a minimal autobahn websocket server example from which to start testing and addressing issues for https://github.com/actix/actix-web/issues/1006.

I assume this example needs to be tweaked slightly to make a reasonable benchmark. From there I can run the autobahn test suite and document all the failures. Then we can pick them off as we see fit.

### DIFF
--- a/websocket-autobahn/config/fuzzingclient.json
+++ b/websocket-autobahn/config/fuzzingclient.json
@@ -7,6 +7,9 @@
     }
   ],
   "cases": ["*"],
-  "exclude-cases": [],
+  "exclude-cases": [
+    "12.*",
+    "13.*"
+  ],
   "exclude-agent-cases": {}
 }
--- a/websocket-autobahn/config/fuzzingclient.json
+++ b/websocket-autobahn/config/fuzzingclient.json
@@ -1,0 +1,12 @@
+{
+  "outdir": "./reports/servers",
+  "servers": [
+    {
+      "agent": "actix-web-actors",
+      "url": "ws://host.docker.internal:9001"
+    }
+  ],
+  "cases": ["*"],
+  "exclude-cases": [],
+  "exclude-agent-cases": {}
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,7 @@ members = [
    "unix-socket",
    "web-cors/backend",
    "websocket",
+   "websocket-autobahn",
    "websocket-chat",
    "websocket-chat-broker",
    "websocket-tcp-chat",
--- a/websocket-autobahn/Cargo.toml
+++ b/websocket-autobahn/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "websocket-autobahn"
+version = "2.0.0"
+authors = ["Mark Lodato <mlodato517@gmail.com"]
+edition = "2018"
+
+[[bin]]
+name = "websocket-autobahn-server"
+path = "src/main.rs"
+
+[dependencies]
+actix = "0.9.0"
+actix-web = "2.0.0"
+actix-web-actors = "2.0.0"
+actix-rt = "1.0.0"
+env_logger = "0.7"
--- a/websocket-autobahn/README.md
+++ b/websocket-autobahn/README.md
@@ -13,27 +13,21 @@ cargo run --bin websocket-autobahn-server
 
 ### Running Autobahn Test Suite
 
-Follow the steps from the [autobahn-testsuite repo](https://github.com/crossbario/autobahn-testsuite#using-the-testsuite-docker-image)
-to install the necessary prerequisites.
+Running the autobahn test suite is easiest using the docker image
+as explained on the [autobahn-testsuite repo](https://github.com/crossbario/autobahn-testsuite#using-the-testsuite-docker-image).
 
-For example:
-
-```bash
-virtualenv ~/wstest
-source ~/wstest/bin/activate
-pip install autobahntestsuite
-```
-
-Once prerequisites are installed, run the test suite:
+First, start a server (see above). Then, run the test suite in fuzzingclient mode:
 
 ```bash
-# In one window, start the server
-cd examples/websocket-autobahn
-cargo run --bin websocket-autobahn-server
-
-# In another window, start the tests
-cd ~
-mkdir test
-cd test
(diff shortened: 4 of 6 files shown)


robjtede:
This is a great start. Pretty handy to have this examples repo around for test server implementations.

I'd rather not require installing the client through pip and virtualenv but use the docker image they provide.

Being on macOS, I've had success running tests using this set up:

`config/fuzzingclient.json`

```json
{
  "outdir": "./reports/servers",
  "servers": [
    {
      "url": "ws://host.docker.internal:9001"
    }
  ],
  "cases": ["*"],
  "exclude-cases": [],
  "exclude-agent-cases": {}
}
```

```sh
docker run -it --rm \
    -v "${PWD}/config:/config" \
    -v "${PWD}/reports:/reports" \
    --network host \
    --name autobahn \
    crossbario/autobahn-testsuite \
    wstest \
    --spec /config/fuzzingclient.json \
    --mode fuzzingclient
```

though on Linux it is probably more appropriate to use `--network host` and a 127.0.0.1 as the URL.

mlodato517:
> I'd rather not require installing the client through pip and virtualenv but use the docker image they provide.

This is a great point! It's what [the docs](https://github.com/crossbario/autobahn-testsuite#using-the-testsuite-docker-image) suggest anyway:
> The following recipe still works, but the new, recommended way is using a Docker toolchain image we provide.

I just have no Docker experience so I avoided it (which is silly).

But I guess is there any reason for me to specify _any_ of this in our README? Should I just link to their README and let anyone who wants to run the test follow the docs the way they prefer? I did just notice that their only docker example is for testing a client, not a server, so I do feel like your example is valuable though 🤔 Maybe it should be added to their README? What do you think?

robjtede:
> is there any reason for me to specify any of this in our README?

I don't see why not. I'd even include the config and reports directories to make it instant for anyone to clone the repo and run the conformance tests without having to look through autobahn docs.

mlodato517:
@robjtede Added [two commits](https://github.com/actix/examples/pull/331/files/6a285ef43a423822e758d0a123d5ba6243a5b4be..f7901d03a8459907364bb62b6f2f6796ad97394b) to hopefully incorporate what you were looking for. Let me know if that's heading in the right direction!

robjtede:
Yep, I like this approach much better. One more tiny detail though, since the reports folder is not necessary we should track a `reports/.gitkeep` file so folks don't have to mkdir it.

mlodato517:
> One more tiny detail though, since the reports folder is not necessary we should track a `reports/.gitkeep` file so folks don't have to mkdir it.

I almost did that but it seemed to me that `wstest` creates the directory if it doesn't exist - do you see an error instead?

robjtede:
> do you see an error instead

No but I can't speak for behavior on Windows or Linux.

mlodato517:
> No but I can't speak for behavior on Windows or Linux.

Safety first! Should I put a `.gitkeep` in `reports/servers` or should we assume `wstest` creates _that_ directory on all systems?

robjtede:
We can be sure it'll write to a dir called reports but I'd feel better having a .gitkeep around since we have the config folder there already.

mlodato517:
Should there be any `.gitignore` updates to ignore reports that get put there? Or do we want those source controlled? Or do we want to rely on people to not add them?

robjtede:
Good idea. A gitignore in that folder set to ignore everything except the .gitkeep would be good too. Probably don't want these tracked ever.

mlodato517:
> Good idea. A gitignore in that folder set to ignore everything except the .gitkeep would be good too. Probably don't want these tracked ever.

Hmmm, sorry to bother you but what's the best way to do that? Should it be:
```
// reports/.gitignore
*
!.gitignore
!.gitkeep
```
or
```
// websocket-autobahn/.gitignore
reports/*
!reports/.gitkeep
```
or some other thing?

robjtede:
np, now that I think about it a gitignore inside the reports folder would serve the same purpose as .gitkeep. File should look then look like this in reports:

```
*
!.gitignore
```

websocket-autobahn/config/fuzzingclient.json
```diff
@@ -0,0 +1,12 @@
+{
+  "outdir": "./reports/servers",
+  "servers": [
+    {
+      "agent": "actix-web-actors",
+      "url": "ws://host.docker.internal:9001"
+    }
+  ],
+  "cases": ["*"],
+  "exclude-cases": [],
```
mlodato517, line 10: Oh, I also meant to ask - are there any tests we should be excluding by default? Not sure if `actix` is already supporting/planning-on-supporting per-message deflate and whatever else

robjtede: Might as well for now. Looks like it's `"12.*"` and `"13.*"`.

mlodato517: 51201fa807600c3bb0f68c4f06c0dc08cd073f01

robjtede:
Great work! I don't have the bandwidth right now to tackle the failures since I'm trying to get v3 in a ready state so if you feel like investigating we'd appreciate it 😄.

mlodato517:
> so if you feel like investigating we'd appreciate it 😄.

First of all - thanks for all the help with this! Secondly, definitely! The whole point of this PR was so that I could create a list to make other PRs off of to fix the other issues :-). I'll run the suite and document the errors on https://github.com/actix/actix-web/issues/1006 and then probably do 1 PR per issue or whatever's most convenient. I just need some help figuring out how to do that. Basically I need help linking to a local crate so I can update, say `actix-web` and then test that it fixed the issue via `actix:examples` or whatever.

robjtede:
First of all, no problem at all. I'm usually responsive on [Gitter]()https://gitter.im/actix/actix-web or the [unoffical Discord](https://discord.gg/2A9h75V) if you want to ask Qs outside of issues. Thanks for the offer to look into this stuff.

Should be able to switch out the dependency on this project with something like `actix-web[-actors] = { path = "../actix-web" }` and then edit the actix-web repo files without any other changes. (You may still run into the usual issues where a `cargo clean` or `rm Cargo.lock` fixes 99% of the time.)

PR #335, opened 2020-06-23
send the correct exit code when closing the websocket

BartWillems:
Fixes https://github.com/actix/actix-web/issues/1555

### DIFF
--- a/websocket-autobahn/src/main.rs
+++ b/websocket-autobahn/src/main.rs
@@ -25,6 +25,10 @@ impl StreamHandler<Result<ws::Message, ws::ProtocolError>> for WebSocket {
                 ws::Message::Text(text) => ctx.text(text),
                 ws::Message::Binary(bin) => ctx.binary(bin),
                 ws::Message::Ping(bytes) => ctx.pong(&bytes),
+                ws::Message::Close(reason) => {
+                    ctx.close(reason);
+                    ctx.stop();
+                }
                 _ => {}
             }
         } else {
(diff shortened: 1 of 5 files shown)


BartWillems:
@fMeow I've updated the styles

websocket-autobahn/src/main.rs
```diff
@@ -25,6 +25,7 @@ impl StreamHandler<Result<ws::Message, ws::ProtocolError>> for WebSocket {
                 ws::Message::Text(text) => ctx.text(text),
                 ws::Message::Binary(bin) => ctx.binary(bin),
                 ws::Message::Ping(bytes) => ctx.pong(&bytes),
+                ws::Message::Close(reason) => ctx.close(reason),
```
robjtede, line 28: a close should be followed by a stop to close the tcp connection

BartWillems: added

mlodato517:
This is a super naiive question, but just for me, when is it "Satisfy autobahn tests with library code" vs "Satisfy autobahn tests with client code"? What's the point of the autobahn test suite? Is it, "Ensure the library _can support_ WS use that satisfies these things", "Ensure the WS code leveraging this library always satisfies these things", or something else?

robjtede:
@mlodato517 The autobhan test suite is a very general way of verfying compliance of a WS server.

It just so happens that the API we have here, in part due to it's flexibility, requires developers to write some of the ping/pong/close/disconnect handling correctly to be compliant, as we found out with this PR.

As we properly analyze each failed test case in actix/actix-web#1006 we will better understand when its library code vs application code. After that, we'll be able to gather and document any requirements that developers should uphold to keep their WS applications spec-compliant. We may even come up with a more ergonomic API to create WS servers, particularly if it seems too onerous on developers to stay compliant using the current system.

mlodato517:
@robjtede Makes sense 👍 I think documenting the user-requirements is a good idea. Not sure if we should document as we go? Also, not sure if the `websocket-autobahn` example could stand as documentation on its own 🤷

PR #336, opened 2020-06-23
Update udp-echo example to actix 0.9 and tokio 0.2.

### LINKED ISSUE #292: Revive some examples
Currently, they're excluded from the workspace:
- [ ] `udp-echo`
- [x] `websocket-chat-broker`

We should bump up the dependencies and revive the examples.

alvardes:
The current `udp-echo` example is based on `actix` `0.7` and `tokio` `0.1` and it is too old to be really useful.

This PR updates the code to be compatible and use the current version of `actix` and `tokio`.

Because of the refactors of the dependencies new dependencies came in:
  * `tokio-util`
  * `futures-util`

Closes #292

### DIFF
--- a/udp-echo/Cargo.toml
+++ b/udp-echo/Cargo.toml
@@ -5,7 +5,10 @@ authors = ["Anton Patrushev <apatrushev@gmail.com>"]
 edition = "2018"
 
 [dependencies]
-actix = "0.7"
-tokio = "0.1"
-futures = "0.1"
-bytes = "0.4"
+actix = "0.9"
+actix-rt = "1.1"
+tokio = "0.2"
+tokio-util = { version = "0.3", features = [ "codec", "udp" ] }
+futures = "0.3"
+futures-util = "0.3"
+bytes = "0.5"
--- a/udp-echo/src/main.rs
+++ b/udp-echo/src/main.rs
@@ -1,42 +1,57 @@
+use actix::io::SinkWrite;
 use actix::{Actor, AsyncContext, Context, Message, StreamHandler};
+use bytes::Bytes;
 use bytes::BytesMut;
 use futures::stream::SplitSink;
-use futures::{Future, Sink, Stream};
+use futures_util::stream::StreamExt;
+use std::io::Result;
 use std::net::SocketAddr;
-use tokio::codec::BytesCodec;
-use tokio::net::{UdpFramed, UdpSocket};
+use tokio::net::UdpSocket;
+use tokio_util::codec::BytesCodec;
+use tokio_util::udp::UdpFramed;
+
+type SinkItem = (Bytes, SocketAddr);
+type UdpSink = SplitSink<UdpFramed<BytesCodec>, SinkItem>;
 
 struct UdpActor {
-    sink: SplitSink<UdpFramed<BytesCodec>>,
+    sink: SinkWrite<SinkItem, UdpSink>,
 }
 impl Actor for UdpActor {
     type Context = Context<Self>;
 }
 
 #[derive(Message)]
+#[rtype(result = "()")]
 struct UdpPacket(BytesMut, SocketAddr);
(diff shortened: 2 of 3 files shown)


udp-echo/Cargo.toml
```diff
@@ -5,7 +5,9 @@ authors = ["Anton Patrushev <apatrushev@gmail.com>"]
 edition = "2018"
 
 [dependencies]
-actix = "0.7"
-tokio = "0.1"
-futures = "0.1"
-bytes = "0.4"
+actix = "0.9"
+tokio = "0.2"
+tokio-util = { version = "0.3", features = [ "udp" ] }
```
Jonathas-Conceicao, line 10: `¨codec¨` feature seams to be missing here

udp-echo/src/main.rs
```diff
@@ -1,42 +1,58 @@
...
-    println!(
-        "Started udp server on: 127.0.0.1:{:?}",
-        sock.local_addr().unwrap().port()
-    );
-
-    let (sink, stream) = UdpFramed::new(sock, BytesCodec::new()).split();
-    UdpActor::create(|ctx| {
-        ctx.add_stream(stream.map(|(data, sender)| UdpPacket(data, sender)));
-        UdpActor { sink: sink }
+    Arbiter::spawn(async {
```
Jonathas-Conceicao, line 40: How about adding `actix-rt` as a dependency and using `actix_rt::main` for an `async` main instead of spawning it as a background future? I think it would be a clearer code and more in tune to how actix apps usually looks like on newer releases.

udp-echo/src/main.rs
```diff
@@ -1,42 +1,58 @@
...
+
 fn main() {
     let sys = actix::System::new("echo-udp");
 
-    let addr: SocketAddr = "127.0.0.1:0".parse().unwrap();
-    let sock = UdpSocket::bind(&addr).unwrap();
-    println!(
-        "Started udp server on: 127.0.0.1:{:?}",
-        sock.local_addr().unwrap().port()
-    );
```
Jonathas-Conceicao, line 33: I think the start message here should also be preserved.

udp-echo/src/main.rs
```diff
@@ -1,42 +1,58 @@
...
+#[rtype(result = "()")]
 struct UdpPacket(BytesMut, SocketAddr);
-impl StreamHandler<UdpPacket, std::io::Error> for UdpActor {
+
+impl StreamHandler<UdpPacket> for UdpActor {
     fn handle(&mut self, msg: UdpPacket, _: &mut Context<Self>) {
         println!("Received: ({:?}, {:?})", msg.0, msg.1);
-        (&mut self.sink).send((msg.0.into(), msg.1)).wait().unwrap();
+
+        self.sink.write(("PING\n".into(), msg.1)).unwrap();
```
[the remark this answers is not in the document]
alvardes: Done.

udp-echo/Cargo.toml
alvardes: Interesting. For me it compiles even if I don't enable codec, even though according the docs it is needed for everything in `tokio_util::codec::*`. I added it.

udp-echo/src/main.rs
alvardes: I tried that, but there is not really a `future` to `await` for. So it shuts down immediately after it starts up.

I changed it to use `actix-rt` and at the end of main it awaits on `actix_rt::Arbiter::local_join()` but I don't think it is a nice solution. Any better idea?

udp-echo/src/main.rs
alvardes: Fixed.

udp-echo/src/main.rs
Jonathas-Conceicao: Being able to make the runtime `await` for internally spawned futures, the `UdpActor` actor in this case, is the intention of the `Arbiter::local_join`. I'm fine with it.

PR #342, opened 2020-07-20
js without jQuery

taisukef:
simplified js code as same functions without jQuery

### DIFF
--- a/websocket-chat/static/websocket.html
+++ b/websocket-chat/static/websocket.html
@@ -1,93 +1,111 @@
 <!DOCTYPE html>
 <html>
-  <meta charset="utf-8" /><head>
+  <head>
     <title>Chat!</title>
-    <script type="module">
+
+    <meta charset="utf-8" />
+
+    <script>
+      'use strict'
+
       window.onload = () => {
-        let conn = null;
+        let conn = null
+
         const log = (msg) => {
-          div_log.innerHTML += msg + "<br>";
-          div_log.scroll(0, div_log.scrollTop + 1000);
-        };
+          div_log.innerHTML += msg + '<br>'
+          div_log.scroll(0, div_log.scrollTop + 1000)
+        }
+
         const connect = () => {
-          disconnect();
+          disconnect()
+
           const wsUri =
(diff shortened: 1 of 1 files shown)


websocket-chat/static/websocket.html
```diff
@@ -1,90 +1,79 @@
...
-      $('#text').keyup(function(e) {
-        if (e.keyCode === 13) {
-          $('#send').click();
-          return false;
-        }
-      });
-    });
+<!DOCTYPE html><html><meta charset="utf-8"><head>
+<title>Chat!</title>
+<script type="module">
```
JohnTitor, line 3: I think this type declaration is unnecessary?

taisukef: Thank you for your review.
if declare type="module", the browser parse as the strict mode for solid program, and the script can use the ES module import.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode

robjtede: The script doesn't use imports and would improve browser compatibility slightly. A 'use strict' declaration would do the same job.

websocket-chat/static/websocket.html
```diff
@@ -1,90 +1,79 @@
...
+window.onload = () => {
+  let conn = null;
+  const log = msg => {
+    div_log.innerHTML += msg + '<br>';
+    div_log.scroll(0, div_log.scrollTop + 1000);
+  };
+  const connect = () => {
+    disconnect();
+    const wsUri = (window.location.protocol === 'https:' && 'wss://' || 'ws://') + window.location.host + '/ws/';
```
robjtede, line 12:
```suggestion
    const wsUri = (window.location.protocol === 'https:' ? 'wss://' : 'ws://') + window.location.host + '/ws/';
```

robjtede:
Thanks for this move off jQuery. Can you run the HTML file through prettier or some similar formatter, too?

robjtede:
hey @taisukef, can you rebase on master and double check this works with v3

robjtede (approved):

PR #347, opened 2020-07-29
Fix rustls example to support pkcs8 keys

ZiiMakc:
By default `mkcert` and I think Let's Encrypt will generate `pkcs8` keys. This PR fix default key usage and ads tip for `rsa` keys.


robjtede:
hey @ZiiMakc , can you rebase onto master and make sure this still works so we can get it merged

robjtede (approved):

PR #355, opened 2020-08-22
Fix for build failure in diesel example

### LINKED ISSUE #354: diesel example fails to build
While attempting to build the diesel example, the build fails with:
```
error[E0433]: failed to resolve: could not find `__rt` in `quote`
   --> /home/shane/.cargo/registry/src/github.com-1ecc6299db9ec823/failure_derive-0.1.6/src/lib.rs:107:70
    |
107 | fn display_body(s: &synstructure::Structure) -> Result<Option<quote::__rt::TokenStream>, Error> {
    |                                                                      ^^^^ could not find `__rt` in `quote`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0433`.
error: could not compile `failure_derive`.

To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...
error: build failed
```

From Googling, this appears to be an issue with the `failure` dependency. See https://github.com/rust-lang-nursery/failure/issues/342 for reference.

I'll submit a PR shortly with a fix.

thesurlydev:
Closes #354

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -16,6 +16,7 @@ bytes = "0.5"
 diesel = { version = "^1.1.0", features = ["sqlite", "r2d2"] }
 dotenv = "0.15"
 env_logger = "0.7"
+failure = "0.1.8"
 futures = "0.3.1"
 r2d2 = "0.8"
 serde = { version = "1.0", features = ["derive"] }


JohnTitor:
Well, I don't think it should be added here as some may be confused by it, "why is there an unused dependency?". The failure v0.1.8, which has a fix for it, was released in May and the users should get the newest version in the future so it won't be a major issue here, I assume. I'd like to get rid of it or add a comment that explains the issue at least.

robjtede:
It will be removed soon when we upgrade examples to v3 anyway.

PR #360, opened 2020-09-03
Fix find_user_by_uid's doc comment

Camotubi:
The previous doc comment was referring to the insert_new_user function


robjtede (approved):
Good catch, thanks 👍

PR #363, opened 2020-09-10
Improve json_decode_error example.

Neopallium:
Show how to use a global `JsonConfig` in the json_decode_error example.


robjtede:
I guess same here, lets just drop the "per-type" syntax.

Neopallium:
@robjtede Done.

robjtede (approved):

PR #364, opened 2020-09-12
v3 examples

robjtede:
Not able to update yet:
- websocket-chat-broker: due to `actix-broker`


redis-session/Cargo.toml
```diff
@@ -5,14 +5,14 @@ authors = ["Nikolay Kim <fafhrd91@gmail.com>", "dowwie <dkcdkg@gmail.com>"]
 edition = "2018"
 
 [dependencies]
-actix-rt = "1.0.0"
-actix-web = "2.0.0"
-actix-session = "0.3.0"
-actix-redis = { version = "0.8.0", features = ["web"] }
+actix-web = "3"
+actix-rt = "1"
```
popzxc, line 9: It seems that `actix-rt` is used for the test only. Shouldn't it be moved to the `dev-dependencies` section?

PR #367, opened 2020-09-14
migrate ci to github actions

robjtede:


.github/workflows/clippy-fmt.yml
```diff
@@ -0,0 +1,71 @@
...
+
+  clippy:
+    name: clippy check
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: cache cargo registry
+        uses: actions/cache@v1
```
JohnTitor, line 36: nit: We could use actions/cache@v2 which has some (performance) improvements.

robjtede:
woo! 3 min checks, think it worked

json_decode_error/src/main.rs
```diff
@@ -34,9 +34,8 @@ async fn main() -> std::io::Result<()> {
     HttpServer::new(|| {
         App::new()
             .service(greet)
-            .app_data(web::JsonConfig::default()
-              .error_handler(json_error_handler)
-            ) // <- register error_handler for JSON extractors.
+            .app_data(web::JsonConfig::default().error_handler(json_error_handler))
+        // <- register error_handler for JSON extractors.
```
JohnTitor, line 38: Should be?

```suggestion
        // ^- register error_handler for JSON extractors.
```

.github/workflows/clippy-fmt.yml
```diff
@@ -0,0 +1,71 @@
...
+          restore-keys: |
+            cargo-index-clippy-${{ matrix.version }}-
+
+      - name: cache cargo target
+        uses: actions/cache@v2
+        with:
+          path: ./target/debug
+          key: cargo-registry-clippy-${{ matrix.version }}-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            cargo-registry-clippy-${{ matrix.version }}-
```
JohnTitor, line 57: We could minify these configs like:
```yaml
path: |
  ~/.cargo/git
  ~/.cargo/registry
  ...
```
as we use the same hashes. But the current is okay if you want to separate restore keys.

JohnTitor (approved):
Overall looks good to me, leaving some nits.

.github/workflows/clippy-fmt.yml
robjtede: We'll have a look, didn't know you could combine them.

robjtede:
I'll be keeping an eye on that cache size. There's some neat tools to reduce it if it becomes a problem.

PR #369, opened 2020-09-18
Removing namespace usage for clarity

emmanuel-keller:


robjtede:
If you're up for it we can do 1 better here and use Data::new and .app_data to prevent a double Arc. Also would need handler extractors changing from `Data<Arc<T>>` to `Data<T>`.

JohnTitor:
Hey @emmanuel-keller, could you rebase onto master so that we could drop the Cargo.lock change and fix CI?

robjtede (approved):

PR #370, opened 2020-09-19
Correct some items' name to fix the build

JohnTitor:


JohnTitor:
Hm, 2.0.0-alpha.7 breaks the build at all, let's fix it actually then.

PR #371, opened 2020-09-26
add Cors to get juniper example to work.

kczimm:
Added actix_cors to get Juniper example to work. The example wouldn't work without it for me. I'm on macOS Catalina using Firefox and Safari browsers. The example would fail with a network error.

### DIFF
--- a/juniper/Cargo.toml
+++ b/juniper/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2018"
 
 [dependencies]
 actix-web = "3"
-actix-cors = "0.3.0"
+actix-cors = "0.4.0"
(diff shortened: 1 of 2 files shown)


juniper/Cargo.toml
```diff
@@ -6,6 +6,7 @@ edition = "2018"
 
 [dependencies]
 actix-web = "3"
+actix-cors = "0.3.0"
```
robjtede, line 9: 0.4.0 was released today

```suggestion
actix-cors = "0.4.0"
```

robjtede (changes requested):

robjtede (approved):

PR #383, opened 2020-11-05
Upgrade websocket-chat-broker example to actix V3 and actix-broker 3.1

cirm:
Upgrading example to actix v3 as dependent actix-broker now can handle the move.


JohnTitor (approved):
Thanks!

PR #389, opened 2020-11-30
Use `rust-argon2` instead of less-maintained `argonautica`

JohnTitor:
`argonautica` is unmaintained and we currently depend on it by a commit hash. We could replace it with an alternative crate to maintain, I'd introduce `rust-argon2` (https://crates.io/crates/rust-argon2). That crate is well-used and enough to demonstrate, I think. This drops some old crates from our deps 🎉 

cc @mygnu, as this will affect your post.


mygnu:
Sounds good to me, I'll update the post on the weekend.

PR #390, opened 2020-12-11
fix typo and update some deps

yjhmelody:


JohnTitor:
Thanks for the PR! Could you drop 0ec7959 to reduce the scope of this PR? And dependency updates are helpful but it's not advisable to make many updates at once. Moreover, some e.g. `futures` are unnecessary here, `cargo update -p` is just fine.

robjtede:
Agreed with the above on updating patch releases on futures. I'd also like to remove the tokio 0.3 updates since it's ill advised at this time to suggest 0.3 works with actix even if these particular examples do. Everything else seems fine.

yjhmelody:
I agree with you !

robjtede (approved):

PR #392, opened 2020-12-16
Refactor template yarte example for use bytes implementation

botika:
Like title

### DIFF
--- a/template_yarte/src/main.rs
+++ b/template_yarte/src/main.rs
@@ -4,21 +4,22 @@ use actix_web::{
     error::ErrorInternalServerError, get, middleware::Logger, web, App, Error,
     HttpResponse, HttpServer,
 };
-use yarte::TemplateMin;
-
-#[derive(TemplateMin)]
-#[template(path = "index")]
-struct IndexTemplate {
-    query: web::Query<HashMap<String, String>>,
-}
+use yarte::ywrite_min;
 
 #[get("/")]
 async fn index(
     query: web::Query<HashMap<String, String>>,
 ) -> Result<HttpResponse, Error> {
-    IndexTemplate { query }
-        .call()
-        .map(|body| {
+    // TODO: check this capacity
+    let mut body = web::BytesMut::with_capacity(4096);
+    // `ywrite_min` is work in progress check your templates before put in production or use `ywrite_html`
+    let mut wrapper = || {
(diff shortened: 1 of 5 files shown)


robjtede:
why the downgrade of yarte from 0.13 to 0.12

botika:
Because yarte 0.13 is provisional to give compatibility has `tokio@0.3` and `bytes@0.6`. 
In yarte 0.14 it will have support for both, but I want to get the new passer out sooner. 
The implementation of 0.12.2 and 0.13 are equal except for the `bytes` version.

template_yarte/src/main.rs
```diff
@@ -1,29 +1,28 @@
...
-        .map(|body| {
-            HttpResponse::Ok()
-                .content_type("text/html; charset=utf-8")
-                .body(body)
-        })
-        .map_err(|_| ErrorInternalServerError("Some error message"))
+    // TODO: check this capacity
+    let mut body = web::BytesMut::with_capacity(4096);
+    // `ywrite_min` is work in progress check your templates before put in production or use `ywrite_html`
+    ywrite_min!(body, "{{> index }}");
```
robjtede, line 21: so, this still ends up using `query` correctly?

template_yarte/src/main.rs
```diff
@@ -1,29 +1,28 @@
...
 ) -> Result<HttpResponse, Error> {
-    IndexTemplate { query }
-        .call()
-        .map(|body| {
-            HttpResponse::Ok()
-                .content_type("text/html; charset=utf-8")
-                .body(body)
-        })
-        .map_err(|_| ErrorInternalServerError("Some error message"))
+    // TODO: check this capacity
```
robjtede, line 18: address todo?

botika: Now I drop this, I forgotten.

template_yarte/src/main.rs
botika: yes, in short what it does is copy code inside a `{}` block. It can have any input and any propagation `?`.

In the template `index.hbs` uses `query.get(...)` any `query` that has `get` method is valid. It's tested :)

robjtede: nice

robjtede (approved):

PR #393, opened 2020-12-18
Improve the "basics" example

neoneye:
- Fixed broken link in readme.
- Removed pixel unit from `` tag. 
- Removed slashes in `` tag and `<link>` tag.


robjtede (approved):
nice, thanks 👍🏻

Issue #348, opened 2020-07-31
tonic example


GopherJ:
Hi, is there anyone using actix-web with tonic?

PR #404, opened 2021-02-01
WIP: add an example of a bidirectional gRPC call via tonic

lunar-debian:
Hi! This is an attempt to add an example of using actix and tonic together (#348) by integrating a client bidirectional gRPC call to other actors.

The proto provides an `Echo` service with an `Echo` RPC. It’s a bidirectional stream, sending `EchoRequest` and received `EchoReply`. I haven’t written the corresponding server yet, but it should be easy enough.

Anyway, this doesn’t build at the moment and I’m not sure to be on the right track…

The current idea is to have an `EchoService`, looking like gRPC, responsible for maintaining a connection to the gRPC server.
It accepts a `RunEcho` message to call the RPC and start an `EchoRpc` actor that would handle input/output for the bidirectional stream. Does it make sense to architecture things this way?

Where I’m stuck is how to implement the actual call to the RPC (see the `RunEcho` handler). I can’t find the proper way to integrate the call without lifetime issues. Any idea?

I'm quite new to Rust, so I would be thankful for any other comments.

This PR is also missing a `README`, which I would happily write after a validation of the general architecture by someone more knowledgeable with Actix and the general ecosystem.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,7 @@ members = [
    "template_tinytemplate",
    "template_yarte",
    "todo",
+   "tonic-bidirectional",
    "udp-echo",
    "unix-socket",
    "web-cors/backend",
--- a/tonic-bidirectional/Cargo.toml
+++ b/tonic-bidirectional/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "tonic-bidirectional"
+version = "0.1.0"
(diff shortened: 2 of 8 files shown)


robjtede:
If there's still interest in getting this PR in a state to merge we'd be interested. Otherwise, if there's no updates, I'll close it out in a couple weeks or so.

PR #406, opened 2021-02-04
Use best practices and remove unecessary unwraps

SuperFluffy:
To guide newcomers on the right path, examples shouldn't unnecessarily use `unwrap`. All functions returning `Result` already have the correct error type, so let's use the question mark operator there.


JohnTitor:
Hey, sorry for the delay, do you have time to fix the build error and rebase on top of master?

PR #407, opened 2021-02-05
Fixes Incorrect Return of body in multipart example

Daggy1234:
The examples shown simply return the html as the body, without setting the right content-type. this results in no html being rendered, but rather just text.


JohnTitor:
Looks reasonable to me, could you resolve the merge conflicts?

Daggy1234:
Sure!

Daggy1234:
Oh looks lik

> Looks reasonable to me, could you resolve the merge conflicts?

Done

JohnTitor:
Great, could you also format code with stable rustfmt?

JohnTitor (approved):
Thanks!

PR #414, opened 2021-03-01
Add tests to diesel example

danitrod:
To facilitate the life of new diesel users who use TDD

_Closes #409_

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: ['stable', 'nightly']
+        version: ["stable", "nightly"]
 
     steps:
       - uses: actions/checkout@v2
@@ -50,11 +50,11 @@ jobs:
           redis-version: 6
 
       - name: run diesel migrations
-        uses: danitrod/diesel-cli-action@v0.3.4
-        with:
-          migration_directories: database_interactions/diesel
-          database_url: test.db
-          sqlite: true
+        run: |
+          cargo install diesel_cli
+          cd database_interactions/diesel
+          diesel migration run
+          chmod a+rwx test.db
 
       - name: cargo test
         uses: actions-rs/cargo@v1
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -49,6 +49,12 @@ jobs:
         with:
           redis-version: 6
 
+      - name: run diesel migrations
+        uses: danitrod/diesel-cli-action@v0.2.0
+        with:
+          migration_directories: database_interactions/diesel/migrations
+          database_url: test.db
+
       - name: cargo test
         uses: actions-rs/cargo@v1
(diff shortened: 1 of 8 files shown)


danitrod:
hmm CI test fails cause there is no sqlite.. I'll try to work on something

danitrod:
ok, managed to do it with a custom github action to run diesel migrations, hope its ok to add another agent to the workflow

.github/workflows/linux.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: ["stable", "nightly"]
+        version: ['stable', 'nightly']
```
JohnTitor, line 18: Why did you change this?

.github/workflows/linux.yml
```diff
@@ -49,6 +49,13 @@ jobs:
         with:
           redis-version: 6
 
+      - name: run diesel migrations
+        uses: danitrod/diesel-cli-action@v0.3.4
+        with:
+          migration_directories: database_interactions/diesel
+          database_url: test.db
+          sqlite: true
```
JohnTitor, line 57: Looking at https://github.com/ducharmemp/diesel-cli-action, I don't think it's complicated enough to use an external action. I'd like to hear some thoughts from @robjtede.

robjtede: yes, please inline this action

.github/workflows/linux.yml
danitrod: Sorry, I think my prettier changed it automatically, will revert

.github/workflows/linux.yml
danitrod: Alright, will try to inline it later today

robjtede (changes requested):

database_interactions/diesel/README.md
```diff
@@ -26,11 +26,18 @@ brew install sqlite3
 cd examples/diesel
 cargo install diesel_cli --no-default-features --features sqlite
 
-echo "DATABASE_URL=test.db" > .env
+echo "DATABASE_URL=prod.db" > .env
```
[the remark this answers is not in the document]
danitrod: Though of using a separate db for tests and prod, then if doing that the test db should be the one called `test.db` - does that make sense?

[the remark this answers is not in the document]
robjtede: I think given that the db files don't exist there is no difference, the test database in CI is ephemeral. Also don't want to give the impression to readers that this example is suitable for production use, as sqlite often isn't.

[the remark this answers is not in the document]
danitrod: true, I was thinking more as kind of a good practice, but you have a point, this shouldn't be seen as production use. i'll use the same db for tests then

robjtede (approved):

JohnTitor (approved):
LGTM, thanks!

PR #415, opened 2021-03-02
Fix sqlx_todo CI checks

danitrod:
Switches the example to use sqlite and adds sqlite3 to the CI.

_Closes #412_

### DIFF
--- a/.github/workflows/clippy-fmt.yml
+++ b/.github/workflows/clippy-fmt.yml
@@ -49,6 +49,14 @@ jobs:
           toolchain: stable
           components: clippy
           override: true
+      
+      - name: create test db for sqlx
+        run: |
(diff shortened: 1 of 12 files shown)


robjtede:
since this adds sqlx back to the workspace, i'd ask that you make clippy happy while we're here :)

danitrod:
My bad, forgot about that. All good now

JohnTitor (approved):
We still have a mention to postgresql on README (Line 3), it should be replaced as well. Otherwise LGTM 👍

danitrod:
nice catch, good to go now :)

JohnTitor:
We've encountered https://github.com/bheisler/TinyTemplate/pull/17, let's just ignore it.
Thanks for fixing it!

PR #421, opened 2021-04-25
Bump tera version to 1.8.0, print the listening port to cli

netroby:


JohnTitor:
You should update the lockfile as well.

netroby:
ok, but it seems will be override by cargo auto , it often changes.

JohnTitor:
> ok, but it seems will be override by cargo auto , it often changes.

I'm not sure what you mean, CI uses the lockfile and it won't be overridden unless we do it by hand.

JohnTitor (approved):

PR #430, opened 2021-06-02
Fix tcp-chat-client(rust) to work

SeanKim:
The `examples/websocket/tcp-chat` crate has two binaries, one for the server and one for the client.
However, there is an error in the client-side code that prevents them from communicating with each other.
This PR corrected these errors and also removed the websocket communication related code that was not directly related to tcp-chat feature and made the client code bloated.


websockets/tcp-chat/src/main.rs
```diff
@@ -228,26 +13,7 @@ async fn main() -> std::io::Result<()> {
...
-            // redirect to websocket.html
-            .service(web::resource("/").route(web::get().to(|| {
-                HttpResponse::Found()
-                    .header("LOCATION", "/static/websocket.html")
-                    .finish()
-            })))
-            // websocket
-            .service(web::resource("/ws/").to(chat_route))
-            // static resources
-            .service(fs::Files::new("/static/", "static/"))
```
JohnTitor, line 248: Since we don't serve `/static` now, we can remove it.

JohnTitor (commented):
Featuring the TCP work (i.e. removing web client) is a good idea to me, to reduce duplication. We should also update README since it's now totally oudated.

websockets/tcp-chat/src/main.rs
SeanKim: thanks for the review! i forgot that part I'll fix it as soon as I get home

websockets/tcp-chat/README.md
```diff
@@ -1,13 +1,12 @@
...
-Added features:
+features:
 
-* Browser WebSocket client
-* Chat server runs in separate thread
-* Tcp listener runs in separate thread
+* Tcp Chat Server
+* Tcp Chat Client
 
 ## Server
```
JohnTitor, line 12: `/name name` below should be removed as well.

robjtede:
I updated this example to work in the v4 migration, if you still thing that this example should remove the websocket part then feel free to update the PR. I'll close this out after a week if there's been no updates.

PR #433, opened 2021-06-07
[EZ] Update comment with correct URL.

jrwats:
http://localhost:8080/ws/index.html gives a 404.  The running server serves `index.html` for the page, and connects to the websocket on `/ws`.

See below:
https://github.com/actix/examples/blob/c3407627d0e45ee8883b1c7d1dc84251ecc6d9a1/websockets/websocket/src/main.rs#L108-L110


robjtede (approved):
Good catch thanks

PR #434, opened 2021-06-13
added mongodb example

douglasOlv:
Hello, Maintainer,
I added mongodb implementation example on `database_interactions/mongodb`, for create a user and list all users.
Lets me know if changes are required. Thanks.

### DIFF
--- a/database_interactions/mongodb/src/main.rs
+++ b/database_interactions/mongodb/src/main.rs
@@ -1,0 +1,47 @@
+use actix_web::{App, HttpServer};
+use dotenv::dotenv;
+use mongodb::{options::ClientOptions, Client};
+use service::UserService;
+use std::sync::Arc;
+
+mod handlers;
+mod service;
+
+struct MongoDB {
+    client: Client,
+}
+impl MongoDB {
+    async fn init() -> Result<Self, mongodb::error::Error> {
+        let client_uri = std::env::var("MONGODB_URI").expect("set MONGODB_URI env var");
+        let client_options = ClientOptions::parse(&client_uri).await?;
+        Ok(MongoDB {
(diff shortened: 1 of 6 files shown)


database_interactions/mongodb/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+}
+
+#[actix_web::main]
+async fn main() -> std::io::Result<()> {
+    dotenv().ok();
+    let mongo = MongoDB::init().await.expect("mongodb connection");
+    let db = mongo.connection_base("myApp");
+    let app_state = Arc::new(AppState {
+        user: UserService::new(db.collection("users")),
+    });
```
thalesfragoso, line 37: Maybe is better to use `web::Data` here instead of `Arc` and `app_data` instead of `data` inside `HttpServer::new`, since Data already uses an `Arc`.

douglasOlv: In this case type it would be  `web::Data<Mutex<my_data>>` ?
 I'm starting my trail with rust, thanks for help.

thalesfragoso: Thank you for the PR. I think you just need to change `Arc::new` to `web::Data::new` and use the `app_data` method instead of the `data` in the `App` type. Let me know if you find any problems.

robjtede:
Thanks for the submission. However, there's nothing really special about the setup of mongo vs the other examples and I don't really want to add things to maintainers plates unnecessarily.

My vote would be to close this PR but would ask contributors for their input on the matter first.

ibraheemdev:
I think it's useful to have a ready-to-go example of mongo, given it is a popular database. See: https://github.com/mongodb/mongo-rust-driver/issues/426.

Issue #445, opened 2021-08-24
`cargo run` in security/rustls panics


efx:
I'm trying to run the rustls example.
Following that README, I 
```sh
cd security/rustls
cargo run
```
 and see the following error:
```sh
Started http server: 127.0.0.1:8443
thread 'main' panicked at 'removal index (is 0) should be < len (is 0)', library/alloc/src/vec/mod.rs:1385:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Panic in Arbiter thread.
```

I will attempt to solve this too.

PR #447, opened 2021-08-24
improve rustls example

efx:
This verifies we have extracted the  private key.

Without this, the example panics with `thread \'main\' panicked at \'removal index (is 0) should be < len (is 0)\', library/alloc/src/vec/mod.rs:1385:13`

Part of #445


security/rustls/src/main.rs
```diff
@@ -47,5 +49,7 @@ async fn main() -> std::io::Result<()> {
     })
     .bind_rustls("127.0.0.1:8443", config)?
     .run()
-    .await
+    .await?;
+    println!("Started http server: 127.0.0.1:8443");
```
robjtede, line 53: this print needs to come before the server await

efx: Ah. I'm curious. Is that because it will not print? I had thought it was misleading printing before the server spawned. I'll certainly move it to after the validation!

robjtede: .await will resolve when the server is shutting down

PR #455, opened 2021-10-02
fix database_interactions/sqlx_todo example

sanchezroselly:
Given 'sqlite' is a prerequisite for this example, 'readme.md' is mod…ified as 'username and password' is not required for 'sqlite'.

File 'model.rs' is also modified in order to amend queries and transactions for 'create' and 'update' functions.


robjtede:
fixed in bfede4c

PR #461, opened 2021-10-22
remove example use async_std::fs

fakeshadow:
It's anti pattern to call any io related `async_std` API inside `actix-web`. 
There is `tokio::fs` that do the same job without extra overhead.


bk138:
Hi guys, I agree that using `async_std` in actix-web is kind of an anti pattern and that this example should be removed in the long term, but removing it as-is leaves the collection without any example showing how to deal with async io. The [multipart example](https://github.com/actix/examples/tree/master/forms/multipart) is nice and all, but one would hardly use this with a production file server. I therefore propose to:

* revert this PR, #461
* add a big fat warning to `multipart-async-std`s README noting that using async-std is actually an anti-pattern and that tokio should be used
* in the long term provide a `multipart-tokio` example and remove `multipart-async-std`

What do you think?

fakeshadow:
example is not meant to show how you write production code. In fact most of the current examples should not be used in production as is.

Adding async-std into actix/example is a no go. multipart example shows you how you handle the file IO correctly inside actix-web. A new example using `tokio::fs` API can be presented to streamline the process but it's the same thing.

bk138:
@fakeshadow I agree with you regarding the proposing of async-std API in actix-examples. 

However, I'd say that examples should indeed show you how to write production code (while not being production-ready, they should at least demonstrate the core concepts...). While `multipart` shows how to do file IO without async-std, it IMO leads users to a sub-par solution as it's delegating blocking IO to a thread pool while doing real async IO would be the better solution.

Anyway, I don't want to be too passionate about this, let's have @robjtede have the last say.

fakeshadow:
There is no real async file IO in async-std. It hide the blocking file IO just like tokio does.

multipart example is as fast and possible faster than any async-std can offer. Because you don't spawn a hidden global runtime and thread pool just for the purpose of hiding blocking file IO.

Like I said a new example including tokio::fs is welcomed. There is nothing to be discussed about bring back async-std stuff. It's not gonna happen.

fakeshadow:
BTW you can get a true user space async file IO with tokio-uring which is already supported by actix-web 4 beta. async-std does not provide anything similar now and in forseeable future.

bk138:
> There is no real async file IO in async-std. It hide the blocking file IO just like tokio does.

Ah OK, didn't know that. Then sorry for the noise.

PR #467, opened 2021-12-10
Fix casbin example for actix-web v3

bk138:
From v2.0.6 on casbin only supports tokio 1.x which conflicts with actix-web 3,
see see https://github.com/casbin/casbin-rs#installation. This commit pins
the casbin dep at v2.0.5 and makes the casbin example work again with current
stable actix-web, i.e. v3.

Please note that once actix-web v4 is out on crates.io, this commit will probably
need to be reverted.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -55,6 +55,8 @@ dependencies = [
  "actix-web 3.3.2",
  "casbin",
  "loge",
+ "rhai",
+ "rhai_codegen",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


robjtede:
Casbin released their breaking tokio v1 change in a patch release?

bk138:
> Casbin released their breaking tokio v1 change in a patch release?

[Yeah, unfortunately](https://github.com/casbin/casbin-rs/releases/tag/v2.0.6).

robjtede (approved):

PR #469, opened 2021-12-27
Upgrade dependencies in websocket-tcp-example

Gentoli:
Update to beta dependencies for `tokio="1"`.

the client panics with `there is no reactor running, must be called from the context of a Tokio 1.x runtime` at `ctx.run_later`

`cargo tree` does not report any `tokio` with version less than 1.


robjtede:
held off because things are being moved over to v4 RC now, if you want to re-use this PR feel free to rebase and update it

robjtede (approved):

PR #471, opened 2022-01-18
Fix deadlock in casbin example

### LINKED ISSUE #470: Casbin example prone to deadlock
The `std::sync::RwLock` used in https://github.com/actix/examples/blob/master/security/casbin/src/main.rs is prone to deadlock. As per https://doc.rust-lang.org/std/sync/struct.RwLock.html#method.write locking the lock might panic if the lock is already held by the current thread. The symptom does not show so much with multiple workers, but is easily triggered with the server limited to 1 worker, please see https://github.com/bk138/examples/blob/casbin-deadlock/security/casbin/src/main.rs for a hacked-up casbin example that [limits the workers to 1 and adds some extra sleep time to one route handler](https://github.com/bk138/examples/commit/9c09cf644e79a0cde5cdb21d7533596714a15e0d). The deadlock can be triggered with this one by doing in the shell:
```
curl http://localhost:8080/success &
curl http://localhost:8080/success &
```
and one gets: 
```
thread 'actix-rt:worker:0' panicked at 'rwlock write lock would result in deadlock', /rustc/02072b482a8b5357f7fb5e5637444ae30e423c40/library/std/src/sys/unix/rwlock.rs:88:13
```

bk138:
...by exchanging the std::sync::RwLock with a
async_std::sync::RwLock. This could also have been the tokio
equivalent but I've used async_std::sync::RwLock as it has nearly the
same API as its std counterpart used in the original casbin example code.

Closes #470

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -57,6 +57,7 @@ dependencies = [
  "loge",
  "rhai",
  "rhai_codegen",
+ "tokio 0.2.25",
 ]
 
 [[package]]
--- a/security/casbin/Cargo.toml
+++ b/security/casbin/Cargo.toml
@@ -13,4 +13,4 @@ casbin = "=2.0.5"
 rhai = "=0.19.14"
 rhai_codegen = "=0.3.3"
 loge = {version = "0.4", default-features = false, features = ["colored", "chrono"]}
-tokio = {version= "*", features = ["sync"]}
+tokio = { version = "0.2.22", features = ["sync"] }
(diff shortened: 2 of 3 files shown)


robjtede:
I would prefer the tokio one so that no deps are added to the tree.

bk138:
> I would prefer the tokio one so that no deps are added to the tree.

Alright, patch updated!

robjtede (approved):

PR #474, opened 2022-01-28
Update basics/basics to v4

LukeMathWalker:
It was pretty straight-forward, with a few exceptions:

- The v4 documentation says that both `Resource` and `Route` can be used as default service, but that's not the case. Only `Route` is accepted;
- `actix-utils` no longer has a `mpsc` module, so I had to work out what adapter was necessary to use `tokio`'s types.


LukeMathWalker:
Related PR on actix-web to fix docs: https://github.com/actix/actix-web/pull/2614

robjtede (approved):

PR #479, opened 2022-01-29
Update basics/state to v4.

barafael:
Issues encountered: deprecated function `App::data` was replaced with
`App::app_data` as suggested in the deprecation notice.


robjtede (approved):
thanks 👍🏻

PR #481, opened 2022-01-29
Updated basics/shutdown-server to v4.

cgubbin:
Moved from `data` to `app_data` in the constructor for `HttpServer`. Rather than cloning the server we now get the handle with `.handle()`.

### DIFF
--- a/basics/shutdown-server/src/main.rs
+++ b/basics/shutdown-server/src/main.rs
@@ -29,7 +29,7 @@ async fn main() -> std::io::Result<()> {
     let server = HttpServer::new(move || {
         // give the server a Sender in .data
         App::new()
-            .data(tx.clone())
+            .app_data(tx.clone())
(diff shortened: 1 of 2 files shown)


basics/shutdown-server/src/main.rs
```diff
@@ -29,7 +29,7 @@ async fn main() -> std::io::Result<()> {
     let server = HttpServer::new(move || {
         // give the server a Sender in .data
         App::new()
-            .data(tx.clone())
+            .app_data(tx.clone())
```
robjtede, line 32: the equivalent to deprecated `.data(val)` is `.app_data(Data::new(val))`

cgubbin: Will fix this.

robjtede (approved):
thanks 👍🏻

PR #482, opened 2022-01-29
Update basics/middleware to v4

LukeMathWalker:
I struggled quite a bit with the redirect middleware due to the requirement of unifying the two different body types (a concrete `BoxBody` with a generic `B` body). 
In the end I opted to use an async block because spelling out the type of `MapOk`, required to call `map_into_boxed_body` on the future returned by `service.call`, was getting out of hand.


robjtede:
@LukeMathWalker updated it with what i'd say are best practices going forwards, hopefully makes sense

LukeMathWalker:
It does! 😁

robjtede (approved):

PR #484, opened 2022-01-29
Update basics/middleware-ext-mut to v4

LukeMathWalker:
I switched `println!` to be `log::info` given that the example initialises a logger.
I also simplified the trait bounds on the middleware itself.


robjtede (approved):
thanks 👍🏻

PR #485, opened 2022-01-29
Update basics/run-in-thread to v4

LukeMathWalker:


robjtede:
arguably this example isn't needed for v4 since we play nice with `#[tokio::main]`; the equivalent now is just `tokio::spawn(server)`

LukeMathWalker:
If the `main` function is async, I agree. But I think there is value in showing how to spawn a server when `main` is synchronous.

PR #487, opened 2022-01-29
Updated basics/middleware-http-to-https to v4.

cgubbin:
I needed to add a new dependency `rustls-pemfile` as the functions for parsing pemfiles are no longer in the core `rustls`.  The `main` function has expanded as the keys and certificates now need to move into `Certificate` and `PrivateKey` structs to be parsed into the `config`.


robjtede (approved):
thanks 👍🏻

PR #498, opened 2022-01-29
Updated forms/form to v4.

cgubbin:
Updated to the new `app_data` method in the constructor. Redid the conversion of response body to `str` as the previous one did not work with the new version's interface.

### DIFF
--- a/forms/form/src/main.rs
+++ b/forms/form/src/main.rs
@@ -23,9 +23,9 @@ async fn main() -> std::io::Result<()> {
 fn app_config(config: &mut web::ServiceConfig) {
     config.service(
         web::scope("")
-            .data(AppState {
+            .app_data(web::Data::new(AppState {
                 foo: "bar".to_string(),
-            })
+            }))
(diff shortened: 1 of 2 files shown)


forms/form/src/main.rs
```diff
@@ -112,12 +102,14 @@ mod tests {
             resp.headers().get(CONTENT_TYPE).unwrap(),
             HeaderValue::from_static("text/plain")
         );
-        assert_eq!(resp.body().as_str(), "Your name is John");
+
+        let body = to_bytes(resp.into_body()).await.unwrap();
+        assert_eq!(body.as_str(), "Your name is John");
     }
 
     #[actix_rt::test]
```
LukeMathWalker, line 110: We can use `actix_web:::test` everywhere (and drop the direct dependency on `actix_rt`).

cgubbin: Thank you! I fixed it.

robjtede (approved):
thanks 👍🏻

PR #506, opened 2022-01-29
Update db/basic, db/sqlx_todo and db/r2d2 to v4

LukeMathWalker:
They had to be updated together due to the shared dependency on the sys crate for sqlite.


robjtede (approved):
thanks 👍🏻

PR #512, opened 2022-01-29
Updated security/rustls to v4.

cgubbin:
Simple update,  changed the parsing of keys/ certs and the configuration builder.


robjtede (approved):
thanks 👍🏻

PR #515, opened 2022-01-31
Updated security/openssl to v4.

cgubbin:
Another simple update, the only change was `header` -> `append_header` in the constructor.


robjtede (approved):
thanks 👍🏻

PR #517, opened 2022-01-31
Updated security/awc_https to v4.

cgubbin:
As `actix_web::client` is no more in v4 I added `reqwest` to make Http requests. I removed `openssl` as a dependancy as `reqwest::Client` implements tls by default.


robjtede:
`actix_web::client` used to be a re-export of `awc` so we should demonstrate usage of the new beta for that crate

cgubbin:
Thank you, I moved everything across to awc.

robjtede (approved):
thanks 👍🏻

PR #521, opened 2022-02-05
Adding auto OpenSSL Let's Encrypt example.

lajtomekadimon:
Here I share a simplified example of how to auto-generate TLS/SSL certificates using OpenSSL with Let's Encrypt on a production server with a real domain. I use acme-micro (based on acme-lib) without using any external tool; I even solve the ACME challenge using Actix itself!

Domain and email values should be edited in main.rs. I didn't know how to provide an example that works anywhere, since this is for a production server. Sorry.


robjtede:
sorry i messed up this branch and can't undo it, superseded in #523

PR #524, opened 2022-02-12
Database interactions/redis

cgubbin:
I updated to V4. I'm not sure if the way I am converting errors from `actix::MailboxError` to `actix_web::Error` is idiomatic, but it works. Other than that everything was simple.


robjtede (approved):
thanks 👍🏻 i cleaned it up a bit, hopefully the tweaks are easy to follow

PR #525, opened 2022-02-13
Updated others/server-sent-events to v4.

cgubbin:
I upgraded to v4. I had to add tokio-stream as a dependency as the stream based stuff used in the example did not make it into tokio v1. I also had to fiddle with the `.js` scripts, as using `localhost` in the http requests didn't work for some reason.


robjtede (approved):
thanks 👍🏻

PR #531, opened 2022-02-22
Yew Server-side Rendering with actix-web Example

zzy:
Yew Server-side Rendering with actix-web Example, use the GitHub REST API.

### DIFF
--- a/yew-ssr-actix-web/README.md
+++ b/yew-ssr-actix-web/README.md
@@ -1,0 +1,8 @@
+# Yew Server-side Rendering with actix-web Example
+
+This example demonstrates yew server-side rendering with actix-web.
+
+This demo need the **development version** of yew.
+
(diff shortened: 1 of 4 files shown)


yew-ssr-actix-web/README.md
```diff
@@ -0,0 +1,8 @@
+# Yew Server-side Rendering with actix-web Example
+
+This example demonstrates yew server-side rendering with actix-web.
+
+This demo need the **development version** of yew.
```
robjtede, line 5: why is the dev version required ?

zzy: SSR would be supported in the Yew Next version, not the current latest version(0.19.3).

robjtede: Ahh that's exciting, we'll wait for its release.

PR #554, opened 2022-06-06
Update databases/postgres/README.md

Raymi306:
Add sudo -u postgres to relevant commands

Add privilege granting step for new user

Add -i flag to curl command to show HTTP response code

I encountered some headaches first time trying to follow the example as a new Postgres user, I hope that these edits will make things more explicit and more likely to work first try without additional troubleshooting or command running.

### DIFF
--- a/databases/postgres/README.md
+++ b/databases/postgres/README.md
@@ -12,7 +12,7 @@
 1. Create database user
 
    ```shell
-   createuser -P test_user
+   sudo -u postgres createuser -P test_user
    ```
 
    Enter a password of your choice. The following instructions assume you used `testing` as password.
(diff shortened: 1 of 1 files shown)


robjtede:
Can you describe more about the environment in which the postgres commands failed? Like OS and how postgres is installed / configured. I suspect that adding the sudo lines will not be a universal solution and may confuse a different set of people? Eg. developing on macOS where there will certainly not be a postgres user.

Raymi306:
It failed on both macOS and on Ubuntu 20.04 Windows Subsystem for Linux version. I wrote this on WSL, I may go back and try and follow it from scratch on the Macbook, but I think the commands were not part of the shell environment there and so I was using PGadmin as a work around

Raymi306:
Checking the macbook again, I installed Postgres with the interactive installer listed first at https://www.postgresql.org/download/macosx/ and don't have any of the shell commands like createuser or createdb as a result, or at least they're not on the path by default.

Raymi306:
@robjtede Are the edits I just committed helpful?

robjtede:
Yep, I think they're more universally helpful. Thanks for the perspective and PR.

PR #557, opened 2022-06-08
Minor lints and clippies

nyurik:
This PR includes #556

* remove unneeded mut
* use explicit `use` instead of the less precise `extern crate` in a simple case
* minor spacing cleanup


robjtede (approved):
nice, thanks

PR #558, opened 2022-06-08
Use captured args in format string

nyurik:
This PR includes #556

Since 1.58, it is possible to use variables directly in the format string, as described in [the blog](https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html#captured-identifiers-in-format-strings)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -47,10 +47,10 @@ dependencies = [
  "log",
  "once_cell",
  "parking_lot 0.11.2",
- "pin-project-lite 0.2.8",
+ "pin-project-lite 0.2.9",
  "smallvec",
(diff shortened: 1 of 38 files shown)


robjtede (approved):
thanks 👍🏻

robjtede:
How'd you do it? I'm guessing there was an automated way of achieving this?

nyurik:
@robjtede search & replace, e.g. for `("[^{"]*)\{\}([^{"]*"), ([a-z_]+)\)` and the like, plus fixing all the other ones by hand (i.e. searching for `\{[:}]` and manual checks).  Some day IDEs will be smarter to do this automagically.

PR #560, opened 2022-06-10
Updated actix-session version to 0.6 in examples/auth/cookie-session

kingledion:
- Updated the actix-session dependency to 0.6.
 - Changed the CookieSession (which is removed in 0.6) to a CookieSessionStore built with MiddlewareSessionBuilder. 
 - The RUST_LOG level was set to  which had the effect of hiding the logs created by the cookie_session (main) application. Changed RUST_LOG to be set to `info` so that all logs are shared.

### DIFF
--- a/auth/cookie-session/Cargo.toml
+++ b/auth/cookie-session/Cargo.toml
@@ -5,6 +5,6 @@ edition = "2021"
 
 [dependencies]
 actix-web = "4"
-actix-session = {version = "0.6", features = ["cookie-session"]}
+actix-session = { version = "0.6", features = ["cookie-session"] }
 log = "0.4"
 env_logger = "0.9"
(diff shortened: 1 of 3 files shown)


auth/cookie-session/Cargo.toml
```diff
@@ -5,6 +5,6 @@ edition = "2021"
 
 [dependencies]
 actix-web = "4"
-actix-session = "0.5"
+actix-session = {version = "0.6", features = ["cookie-session"]}
```
JohnTitor, line 8: nit:
```suggestion
actix-session = { version = "0.6", features = ["cookie-session"] }
```

JohnTitor (commented):
Leaving one nit, could you also run rustfmt?

JohnTitor (approved):
Thanks!

JohnTitor:
Yeah, `cargo fmt` invokes rustfmt and it's a preferable way here.

PR #562, opened 2022-06-24
Replace all unwraps in hello-world example with ? operator

PeterPierinakos:
Hello. This is a pretty insignificant change, but I noticed that the hello-world example uses 2 unwraps in the tests even though it returns a `Result`, so I thought that it should use `?` instead of `.unwrap()`. This PR changes 2 lines of code in the `test_index` test.


robjtede (approved):
thanks

PR #563, opened 2022-07-09
update register link to match correct path

BSPlatt:
There is no registered route to /invitation/{invitation_id}. Instead, it is /register/{invitation_id}


robjtede (approved):

robjtede:
Thanks

PR #574, opened 2022-09-07
remove wrong cd commands

sousajo-cc:
one is already at databases/diesel


sousajo-cc:
@robjtede I was taking a look at the example and just found this.

robjtede:
Due to this repo's setup we don't want to assume what the CWD is, and in this case it's important to be in the specific example's directory so we highlight it on purpose

PR #585, opened 2022-11-21
Fix invalid non-existent link (for database example:rbatis).

u-u-z:
The "rbatis example with actix-web" content in the original documentation is **invalid**

I found a example that shows "how rbatis works with actix" from the documentation of the new version of **rbaits-v4**

https://github.com/rbatis/rbatis/blob/master/Readme.md?plain=1#L30-L31

### DIFF
--- a/databases/rbatis/README.md
+++ b/databases/rbatis/README.md
@@ -1,5 +1,3 @@
 The `rbatis` ORM developers keep their own example on usage with Actix Web.
 
-You can find the example in the [rbatis/example/src/actix_web](https://github.com/rbatis/rbatis/blob/master/example/src/actix_web/main.rs) directory.
-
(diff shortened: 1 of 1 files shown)


JohnTitor:
Seems they removed the actix-web example from the repo, and abs_admin is more than an example, I think.
I'd prefer to get our old example back from the commit history instead.

u-u-z:
> Seems they removed the actix-web example from the repo, and abs_admin is more of showcase than an example, I think. I'd prefer to get our old example back from the commit history instead.

😄 Sounds good ! I found it here: I
https://github.com/rbatis/rbatis/blob/a1c09af45d9827f658e9f7ef5b58419e95b25fe7/example/src/actix_web/main.rs

I will patch it!

robjtede (approved):
Thanks

PR #588, opened 2022-11-23
bugs and warning fixes

ccarnero:
Fixing:

"severity": 8
"message": "cannot move out of `secret_key` a captured variable in an `Fn` closure\nmove occurs because `secret_key` has type `Key`, which does not implement the `Copy` trait",

"severity": 4,
"message": "unused import: `rand::Rng`\n`#[warn(unused_imports)]` on by default"


robjtede (approved):
Thanks 👍

PR #589, opened 2022-12-03
Fixes to autobahn websocket echo server

mskorkowski:
Reimplementation of the autobahn example.

Fixes most of the missing parts of  https://github.com/actix/actix-web/issues/1006


robjtede:
stale, test suite server has been converted to actix-ws

PR #590, opened 2022-12-10
fetch text from multipart/form-data

zonblade:
Adding example for fetching text field in multipart/form-data

- converting to key-value hashmap.


zonblade:
my bad, didn't see the docs.
[this](https://docs.rs/actix-web/4.2.1/actix_web/web/struct.Payload.html) provide to read text payload given.

PR #594, opened 2022-12-21
Use workspace dependencies for actix-web

JohnTitor:
`[workspace.dependencies]` is available since 1.64: https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table
This replaces package-level deps with that for all the actix-related crates, we now can upgrade their versions in one place.

Signed-off-by: Yuki Okushi <jtitor@2k36.org>


robjtede (approved):
great idea

PR #595, opened 2022-12-29
Fix auth/cookie-auth, add time to live to SessionMiddleware

D0bhareach:
Master branch wasn't working index and login requests responded with 405. Added some comments about the Key. Add ttl to SessionMiddleware::build() for educational purposes.

Question: Is it possible to hot rotate Key? Move reading Key in the closure, update keys with chron. The sessions that are valid will invalidate automatically and will need to re-login?

### DIFF
--- a/auth/cookie-auth/src/main.rs
+++ b/auth/cookie-auth/src/main.rs
@@ -1,15 +1,20 @@
 use actix_identity::{Identity, IdentityMiddleware};
-use actix_session::{storage::CookieSessionStore, SessionMiddleware};
+use actix_session::{config::PersistentSession, storage::CookieSessionStore, SessionMiddleware};
 use actix_web::{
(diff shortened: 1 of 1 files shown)


auth/cookie-auth/src/main.rs
```diff
@@ -35,19 +40,32 @@ async fn main() -> std::io::Result<()> {
...
+                .cookie_secure(false)
+                .session_lifecycle(
+                    PersistentSession::default().session_ttl(Duration::seconds(JUST_A_MINUTE)),
+                )
+                .build(),
             )
             // enable logger - always register Actix Web Logger middleware last
             .wrap(middleware::Logger::default())
-            .service(web::resource("/login").route(web::post().to(login)))
+            .service(web::resource("/login").route(web::get().to(login)))
```
robjtede, line 68: this was correct, /login should respond only to POST requests

robjtede:
took some piece from this, improve example in de5e6eecc0a80014c113dffca9a49d06df0b0a0e

PR #599, opened 2023-01-08
Fixed "Method not allowed" on the todo app due to 307 redirect not ch…

### LINKED ISSUE #598: The basic/todo app breaks with 405 (method not allowed) after create/update action applied on an item
This happens due to 307 redirect to "/" not changing HTTP method and the index route listening for GET requests only.

iamruss:
…anging HTTP method.
This PR fixes #598 .


robjtede (approved):
good catch, thanks

PR #603, opened 2023-02-13
Add an example returning HttpResponse from middleware.

zonblade:
Implementation of returning HttpResponse from middleware.


robjtede:
You need to add this example to the root workspace Cargo.toml in order to make sure it's tested in CI.

PR #604, opened 2023-02-19
auth/simple-auth-server: fix panic due to default `SECRET_KEY` being 32 bytes, not 64 bytes

### LINKED ISSUE #605: auth/simple-auth-server panics when default `SECRET_KEY`  is used to instantiate a `actix_web::cookie::Key` instance
Hi,

on URLO someone raised the issue that the simple-auth-server panics:

https://users.rust-lang.org/t/trying-to-launch-auth-web-microservice-with-rust-using-actix-web/89498/1

I took a look and saw that the default [`SECRET_KEY`](https://github.com/jofas/actix-examples/blob/2df944c5e55951021e6c1da0feffef8c24c19506/auth/simple-auth-server/src/utils.rs#L6) used when the user does not set the `SECRET_KEY` environment variable is only 32 byte long. `actix-web v4` requires the key to be 64 byte long: https://docs.rs/actix-web/latest/actix_web/cookie/struct.Key.html#method.from. The `SECRET_KEY` is used [here](https://github.com/jofas/actix-examples/blob/2df944c5e55951021e6c1da0feffef8c24c19506/auth/simple-auth-server/src/main.rs#L47) to build cookie session middleware.

jofas:
Fixes #605

### DIFF
--- a/auth/simple-auth-server/src/utils.rs
+++ b/auth/simple-auth-server/src/utils.rs
@@ -4,7 +4,7 @@ use once_cell::sync::Lazy;
 use crate::errors::ServiceError;
 
 pub static SECRET_KEY: Lazy<String> =
-    Lazy::new(|| std::env::var("SECRET_KEY").unwrap_or_else(|_| "0123".repeat(8)));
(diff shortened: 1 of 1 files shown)


JohnTitor (approved):

JohnTitor:
Thank you!

PR #606, opened 2023-02-25
https-tls: typo - correct logged URIs to use https

ajcollins:
When exploring the https-tls examples, I noticed that the URIs printed to screen were using the wrong protocol.

This PR corrects the protocol to use https, so that if you click on the link it works as expected.


robjtede (approved):
Thank you :)

ajcollins:
Not at all: thank you!

PR #610, opened 2023-03-08
fix test for forms/form

bhbs:
```sh
cd forms/form
cargo test
```

### DIFF
--- a/forms/form/src/main.rs
+++ b/forms/form/src/main.rs
@@ -119,7 +119,7 @@ mod tests {
                 name: "John".to_string(),
             })
             .to_request();
-        let resp: ServiceResponse = app.call(req).await.unwrap();
+        let resp: ServiceResponse = test::call_service(&app, req).await;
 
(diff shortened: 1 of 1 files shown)


forms/form/src/main.rs
```diff
@@ -186,7 +186,7 @@ mod tests {
             name: "John".to_string(),
         });
         let result = handle_post_3(req.clone(), params).await;
-        let resp = result.respond_to(&req);
+        let resp = result.customize().respond_to(&req);
```
robjtede, line 189: why is this necessary? no customization is being done.

bhbs: I'm sorry, this might be a beginner question, but I am getting this error.
How can it be suppressed?

bhbs: Thanks for the review. Then I will leave it untouched in this pull request.

robjtede (approved):

PR #629, opened 2023-07-22
feat: update example api to support GET /users

### LINKED ISSUE #628: Expand postgres example to include querying/fetching all
It looks like the postgres example shows how [inserting new rows](https://github.com/actix/examples/blob/master/databases/postgres/src/main.rs#L59-L81) works, but doesn't cover how you would implement row querying. 

The disel example shows inserts [and queries](https://github.com/actix/examples/blob/master/databases/diesel/src/actions.rs#L9-L21), however. 

Since [the sql is already available](https://github.com/actix/examples/blob/master/databases/postgres/sql/get_users.sql) for querying users, I imagine I can just implement a new handler using the example for inserts. 

If the changes are as straightforward as they appear, I should be able to have a PR done by end of day.

joeyfigaro:
Added support to postgres docs for querying all users using existing get_users.sql. 

Closes #628

### DIFF
--- a/databases/postgres/src/main.rs
+++ b/databases/postgres/src/main.rs
@@ -57,9 +57,9 @@ mod db {
     use crate::{errors::MyError, models::User};
 
     pub async fn get_users(client: &Client) -> Result<Vec<User>, MyError> {
-        let _stmt = include_str!("../sql/get_users.sql");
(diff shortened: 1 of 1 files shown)


robjtede (approved):

robjtede:
after a rustfmt run i'll merge it

PR #630, opened 2023-07-24
Middleware for auto caching succesful GET responses with redis

bangbaew:


bangbaew:
README.md is updated on the base branch

cache/redis-cache-middleware/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+name = "main"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+actix-web = "*"
+actix-rt = "*"
+env_logger = "*"
+actix-web-lab = "*"
+redis = { version = "*", features = ["tls-native-tls"] }
```
robjtede, line 12: using * deps is very bad practice, please copy examples format

for any deps in the workspace root you can use `<dep>.workspace = true`

cache/redis-cache-middleware/src/main.rs
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    for i in 2..=n {
+        print!("{i} ");
+        c = a + b;
+        a = b;
+        b = c;
+    }
+    b
+} */
```
robjtede, line 40: dead code

cache/redis-cache-middleware/src/main.rs
robjtede, line 1: Some tests would be good if you feel like the challenge. We run redis in CI for other examples so should be easy enough to copy those.

robjtede (changes requested):

bangbaew:
added dependency versions and removed dead code

robjtede:
I'd be interested to merge this if we can fix the CI.

JohnTitor (approved):
Thanks!

PR #636, opened 2023-09-14
Fixed return value of index() function

TahaOps:


robjtede:
Until `Html` is moved to Actix Web, we will continue to demonstrate it here.

Your change is not equivalent, but it's an easy mistake to make which is why we're creating a `Html` responder type. To do this without the lab crate:

```rust
Ok(HttpResponse::Ok().content_type(ContentType::html()).body(s))
```

PR #640, opened 2023-10-08
fix python docs and added venv

Thomblin:
I did just test this example. I realized that `pip3 install --requirements requirements.txt` and I suggest to always use a venv when using python to separate dependencies


robjtede (approved):
Thanks for the suggestion 👍🏻

PR #673, opened 2023-12-08
Fix small issues in juniper-advanced example

ethagnawl:
This PR addresses two issues I ran into while running the "juniper-advanced" example:

- Encoding error in mysql-schema.sql
  - I'm admittedly not sure how broadly useful this is, as I don't regularly use mysql but I had to replace instances of `utf8mb` with `utf8` and `utf8mb4_0900_ai_ci` with `utf8_general_ci` in order to import the DB schema. I'm running: `mysql  Ver 15.1 Distrib 10.5.21-MariaDB, for debian-linux-gnu (x86_64) using  EditLine wrapper`
- Fix schema descriptions in root.rs which were referring to the wrong table (`user` => `product`)


robjtede:
> I had to replace instances of utf8mb with utf8

hmm this is definitely not right, `utf8` is legacy and not really correct

robjtede (approved):

robjtede:
I'm guessing the newer utf8mb4_0900_ai_ci collation is not supported in your version of mysql. I've changed the charset back to utf8mb4 and updated the collate to utf8mb4_unicode_ci to support a larger range of mysql versions. Let me know if that combo works for you.

PR #699, opened 2024-01-02
fixed env variable not loading

kunal171:
The code was not running because we had to load env file with dotenv().ok() function to load environment variables


robjtede:
i don't think this generally adds to the example presented

PR #736, opened 2024-02-05
Add community example: a simple blog for markdown rendering 

gemini-15:
Proposition for a community example of a blog showcasing the use of actix with diesel and r2d2 and API endpoints. 
Usage with react.


robjtede:
thanks 👍🏻

PR #744, opened 2024-02-16
Example of using rustls v0.22

ConsoleC137:
Since `actix-web` now supports `rustls v0.22`, I feel it is necessary to show how to use it.


robjtede:
examples have been updated to use v0.23

PR #796, opened 2024-04-15
update(README.md): adds Angular + Actix with JWT security example

stav121:
Hello there,

I am adding my submission, a boilerplate project that contains a JWT implementation on Actix, with Angular frontend and PostgreSQL that is already Dockerized.

Best regards!


robjtede (approved):
nice one, thanks 👍🏻

PR #805, opened 2024-04-30
uploaded example using JWT and Rusqlite for authentication 

Stipulations:
uploaded example using JWT and Rusqlite for authentication


robjtede:
if you're interested in this being accepted still, can you please:
- move it into the auth/ dir
- give it a shorter name
- add a readme (see other examples) with instructions for usage etc.
- instead of a premade DB file, add a .sql file with instructions in the README on how to create a users.db file locally
- .gitignore the users.db file location

PR #863, opened 2024-07-22
Move client file from old echo example to actorless

c-git:
Not sure about this change but decided to create the PR and let you decide instead.

- Make it easier for users to find rust code they can use for the client
- Remove the need to install websocat to be able to use cli
- Soft linked from the old example to the moved file so if the old one gets removed it doesn't break.


robjtede:
thanks 👍🏻

PR #892, opened 2024-08-21
Fix chat-tcp panic when current directory is not the expected directory

JonathanLindsey:
Prior to this change if you followed the websockets/chat-tcp README without first making your current directory websockets/chat-tcp, you would get a panic when you opened http://localhost:8080/ .

This uses `include_str!()` to include the index.html file at compile time referencing it relative to the source code file and avoids the use of the problematic `unwrap()` all together.

### DIFF
--- a/websockets/chat-tcp/README.md
+++ b/websockets/chat-tcp/README.md
@@ -18,7 +18,14 @@ Chat server listens for incoming tcp connections. Server can access several type
 - `some message` - just string, send message to all peers in same room
 - client has to send heartbeat `Ping` messages, if server does not receive a heartbeat message for 10 seconds connection gets dropped
 
-To start server use command: `cargo run --bin websocket-tcp-server`
+To start server run
+
+```sh
+cd websockets/chat-tcp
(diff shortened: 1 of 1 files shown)


robjtede:
Usage of NamedFile here is preferred. We should update the instructions to include a `cd` command.

JonathanLindsey:
Ok will stick with NamedFile. But the problem extends to a deployed executable. If you build the executable and move the `target/release` folder to another location that does not have the source code it won't have `index.html`. Even if you had a copy of `index.html`, the current directory you launch the executable from matters...which I think shouldn't matter *at least in the deployed state*. If you accidentally started the executable from a different directory, you could expose a different `index.html` file that you didn't mean to besides the application being broken.

Would it make since to copy `index.html` to the `target/debug [or] release/static` folder at build time using a `build.rs` file? Then load `index.html` at runtime with a path relative to the executable file rather than the current directory.

JonathanLindsey:
My above idea led me down a few rabbit holes so just adding the instruction to cd to the right directory.

robjtede (approved):

PR #912, opened 2024-09-12
Ractor WebSockets

robjtede:
just demo-ing this actor lib: https://crates.io/crates/ractor

### DIFF
--- a/websockets/echo/src/main.rs
+++ b/websockets/echo/src/main.rs
@@ -4,10 +4,10 @@
 
 use actix_files::NamedFile;
 use actix_web::{middleware, web, App, Error, HttpRequest, HttpResponse, HttpServer, Responder};
-use actix_web_actors::ws;
+use ractor::Actor;
 
 mod server;
-use self::server::MyWebSocket;
+use self::server::{MyWebSocket, WsMessage};
 
 async fn index() -> impl Responder {
     NamedFile::open_async("./static/index.html").await.unwrap()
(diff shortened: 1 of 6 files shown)


websockets/echo/src/server.rs
```diff
@@ -11,68 +10,84 @@ const CLIENT_TIMEOUT: Duration = Duration::from_secs(10);
...
-                ctx.stop();
+
+            AsMessage::Ws(msg) => {
+                // process websocket messages
+                println!("WS: {msg:?}");
+
+                match msg {
+                    AggregatedMessage::Ping(msg) => {
+                        state.0 = Instant::now();
+                        state.1.as_mut().unwrap().pong(&msg).await.unwrap();
```
slawlor, line 68: if killing the actor is intended with the `unwrap()`, you also can just use `?` notation as we capture anything that can be typed to an `Error`. The benefit of that is you maintain the error source in the Error trace.

robjtede: Ahh right, hadn't spotted that ActorProcessingErr was an alias.

websockets/echo/src/server.rs
```diff
@@ -1,78 +1,111 @@
...
-    /// Client must send ping at least once per 10 seconds (CLIENT_TIMEOUT),
-    /// otherwise we drop connection.
-    hb: Instant,
-}
+pub(crate) struct MyWebSocket;
 
 impl MyWebSocket {
-    pub fn new() -> Self {
-        Self { hb: Instant::now() }
+    async fn handle_hb(
```
slawlor, line 22: nit: do you need `&self` at all here? Since everything is mutated on the state, it seems like it would be "cleanest" to have a newtype for the state, and these are mutable fn's on the state.

websockets/echo/src/main.rs
```diff
@@ -4,18 +4,30 @@
...
 }
 
 /// WebSocket handshake and start `MyWebSocket` actor.
 async fn echo_ws(req: HttpRequest, stream: web::Payload) -> Result<HttpResponse, Error> {
-    ws::start(MyWebSocket::new(), &req, stream)
+    let (res, session, stream) = actix_ws::handle(&req, stream)?;
+
+    let (actor, _handle) = Actor::spawn(None, MyWebSocket, session).await.unwrap();
+
+    actix_web::rt::spawn(async move {
```
slawlor, line 22: we also have utilities in a companion crate which adds the actor protection on top of stream handling, so you can benefit from supervision and cleanly handling streams. https://docs.rs/ractor_actors/latest/ractor_actors/streams/pump/fn.spawn_stream_pump.html

but that's just an aside.

slawlor (approved):

websockets/echo/src/server.rs
robjtede: had a go but still need the zst for the actor impl block, so opted to keep these as-is

PR #992, opened 2025-01-21
Diesel-async usage example

alexted:
I think it's useful nowadays to have a code example like this as a starting point.

### DIFF
--- a/databases/diesel-async/Cargo.toml
+++ b/databases/diesel-async/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "db-diesel-async"
+version = "1.0.0"
+edition = "2021"
+
+[dependencies]
+actix-web.workspace = true
+diesel = { version = "*", default-features = false, features = ["uuid"] }
+diesel-async = { version = "*", features = ["postgres", "bb8", "async-connection-wrapper"] }
+serde.workspace = true
+uuid.workspace = true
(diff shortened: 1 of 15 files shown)


databases/diesel-async/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "db-diesel-async"
+version = "1.0.0"
+edition = "2021"
+
+[dependencies]
+actix-web.workspace = true
+diesel = { version = "*", default-features = false, features = ["uuid"] }
+diesel-async = { version = "*", features = ["postgres", "bb8", "async-connection-wrapper"] }
```
robjtede, line 9: please don't use * versions (ever :P)

robjtede:
Yep I'll check this out later.

robjtede:
please feel free to squash and/or perform a rebase/force push on your branch

please try not to update unrelated workspace deps in the future, the conflicts have become very complicated to resolve cleanly as a result

robjtede:
Maybe we could mark as ignored? Alternatively rewrite this to use sqlite which is how we've gotten around needing DBs running in CI with the other examples.

robjtede (approved):
good addition, thanks 👍🏻 

i may make some tweaks once merged so as not to bore you with nits

PR #1082, opened 2025-05-06
refactor: improve select usage with tokio::select! macro in websockets/actor-actorless example

destinyFvcker:
I noticed the TODO comment "nested select is pretty gross for readability on the match" in the example. To address this, I replaced the usage with tokio::select!, which simplifies the control flow and improves readability, especially around pattern matching.

According to [this thread](https://users.rust-lang.org/t/why-doesnt-tokio-select-require-fusedfuture/46975) on the Rust forums, tokio::select! is more ergonomic than futures::select!, as it does not require FusedFuture, pin!, or additional boilerplate.

### DIFF
--- a/websockets/chat-actorless/src/handler.rs
+++ b/websockets/chat-actorless/src/handler.rs
@@ -1,13 +1,7 @@
-use std::{
-    pin::pin,
-    time::{Duration, Instant},
-};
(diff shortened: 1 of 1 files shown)


robjtede (approved):
a lot simpler, thanks 👍🏻

PR #1154, opened 2025-09-05
Update Postgres example to use ThinData

lucascompython:
This PR updates deps, fixes clippy warnings and uses ThinData instead of Data for db pool.
As mentions in [this](https://github.com/actix/examples/discussions/1153#discussioncomment-14287197) discussion, `ThinData` should be used here.


robjtede (approved):
thanks 👍🏻

PR #1298, opened 2026-04-21
chore: drop shuttle-related stuff

JohnTitor:
ref. https://docs.shuttle.dev/docs/shuttle-shutdown
Also https://github.com/sentinel1909/shuttle-templat-actix is now 404.

### DIFF
--- a/auth/oauth-github/README.md
+++ b/auth/oauth-github/README.md
@@ -1,7 +1,0 @@
-# OAuth (GitHub)
-
-See <https://github.com/robjtede/actix-examples-oauth-github>.
-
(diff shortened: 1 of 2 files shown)


auth/oauth-github/README.md
robjtede, line 1: ahhh noo, i've migrated my example off shuttle

forgot about this reference

JohnTitor: Ah, sorry I didn't check it carefully!
Would you mind reviving that?
Otherwise I'll handle it tomorrow.

